PR #2, opened 2023-06-05
mock

Siddharth2207:


thedavidmeister:
this is not how foundry does mocking https://book.getfoundry.sh/cheatcodes/mock-call?highlight=mock#mockcall

PR #42, opened 2023-11-09
Subgraph code base for OrdrrBook Subgraph - Part #1 - SG Tests

NanezX:
First split for the OB Subgraph


NanezX:
Well, basically this provide the source code of OB subgraph. Already have 2.3k, so I think I'm gonna consult with you guys about. This also provide the necessary npm deps for building the SG and also ignoring all the generated files.

Use forge to generate the ABIs and nix for initializing the files required by the SG.

@thedavidmeister @hardyjosh

subgraph/flake.nix
```diff
@@ -0,0 +1,47 @@
+{
+  description = "Flake for development orderbook subgraph workflows.";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
```
thedavidmeister, line 5: let's pick a specific commit

subgraph/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "Rain OrderBook Subgraph",
+  "version": "0.1.0",
+  "author": "rainprotocol",
+  "license": "CAL",
+  "scripts": {
+    "codegen": "graph codegen",
+    "build": "graph build"
+  },
+  "dependencies": {
+    "@graphprotocol/graph-cli": "0.46.1",
```
thedavidmeister, line 12: can we pin our deps?

subgraph/src/utils.ts
```diff
@@ -0,0 +1,384 @@
+import {
```
thedavidmeister, line 1: can we please organise this "utils" file into dedicated files with descriptive names and tests?

subgraph/subgraph.template.yaml
```diff
@@ -0,0 +1,68 @@
...
+schema:
+  file: ./schema.graphql
+dataSources:
+  - kind: ethereum/contract
+    name: OrderBook
+    network: {{ network }}
+    source:
+      address: "{{ orderbook }}"
+      abi: OrderBook
+      startBlock: {{ blockNumber }}
```
thedavidmeister, line 11: can we do yaml manipulation as structured data rather than with string interpolation?

thedavidmeister (changes requested):

subgraph/flake.nix
NanezX: Ok, it's using commit: `ec750fd01963ab6b20ee1f0cb488754e8036d89d`

subgraph/package.json
NanezX: Well, that way also is for pinning deps, always will resolve to that specific version. But I used the `=` to make it explicit.

subgraph/subgraph.template.yaml
NanezX: Ok, I'm using `serde_yaml` to manipulate the subgraph template from rust (i built a cli to do this aswell) and modify each field.

The issue that I had was handling the address present in the YAML file since it was not possible for me to put the address with quotes using just serde_yaml. This because serde_yaml put the address as a hex value without quotes.

I tried using the `format` macro to put the quotes but then extra quotes are added by the serialization - if I use single quotes, then the output give me 3 single quotes, and if I use double quotes, then I get single quote + double quote surrounded the address. 

So, I ended using `replace` on the string before writing the file

subgraph/src/utils.ts
NanezX: Reorganized into dedicated folders and files. Tested respective functions on package repository.

subgraph/src/utils/entity/get.ts
```diff
@@ -0,0 +1,28 @@
+import { Address, Bytes } from "@graphprotocol/graph-ts";
+import { OrderBook, RainMetaV1 } from "../../../generated/schema";
+import { getKeccak256FromBytes } from "@rainprotocol/subgraph-utils";
+
+export function getOB(obAddress_: Address): OrderBook {
```
thedavidmeister, line 5: weird var name with _

subgraph/src/utils/constants/index.ts
```diff
@@ -0,0 +1,15 @@
+// Orderbook: TakeOrder(address sender, TakeOrderConfig config, uint256 input, uint256 output)
```
thedavidmeister, line 1: could these values actually be calculated rather than hardcoded with a comment? what if the comment and hash get out of sync? how will we know?

subgraph/src/orderBook.ts
```diff
@@ -0,0 +1,1166 @@
...
+
+    contextEntity.emitter = contextTakeOrder.emitter;
+    contextEntity.timestamp = contextTakeOrder.timestamp;
+    contextEntity.transaction = contextTakeOrder.transaction;
+
+    takeOrderEntity.context = contextEntity.id;
+
+    contextEntity.save();
+
+    // store.remove("ContextEntity", "ContextTakeOrderTemp");
```
thedavidmeister, line 732: commented out code

subgraph/src/orderBook.ts
```diff
@@ -0,0 +1,1166 @@
...
+        metaContent.contentLanguage = this.contentLanguage;
+    }
+
+    const auxParents = metaContent.parents;
+    if (!auxParents.includes(this.rainMetaId)) auxParents.push(this.rainMetaId);
+    metaContent.parents = auxParents;
+
+    this.metaContent = metaContent;
+    this.metaStored = true;
+    // metaContent.save();
```
thedavidmeister, line 1156: commented code

subgraph/cli/subgraph/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    for template in &mut yaml_data.templates {
+        template.network = args.network.clone();
+    }
+
+    let mut modified_yaml = serde_yaml::to_string(&yaml_data)?;
+
+    // TODO: Modifiy this since when serializing the string does not add the quotes.
+    // And when the quotes are added using format! macro, then two or three quotes
+    // are added.
+    modified_yaml = modified_yaml.replace("'\"", "'");
```
thedavidmeister, line 113: we should not do this

thedavidmeister (changes requested):

subgraph/subgraph.template.yaml
NanezX: I'll mark as resolved since this code is now part of other repo: [subgraph-cli](https://github.com/rainprotocol/rain.subgraph-cli) which is used inside the nix flake,

PR #51, opened 2023-12-23
2023 12 23 ob crate

Siddharth2207:


orderbook-rs/src/gasoracle/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+/// * `chain_id` - Network Chain Id.
+///
+pub async fn gas_price_oracle(
+    api_key: Option<String>,
+    chain_id: u64,
+) -> anyhow::Result<(f64, f64)> {
+    let client = Client::new();
+    let url = format!(
+        "{}{}",
+        "https://api.blocknative.com/gasprices/blockprices?chainid=", chain_id
```
thedavidmeister, line 19: magic strings like this should probably be a constant, especially if we plan to ever hit more than one endpoint for an API

also there is a url crate fyi https://docs.rs/url/latest/url/

orderbook-rs/src/gasoracle/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    );
+    let mut request = client.get(url);
+    if let Some(api_key) = api_key.as_ref() {
+        request = request.header(AUTHORIZATION, api_key);
+    }
+    let response: BlockNativeResponse = request.send().await?.error_for_status()?.json().await?;
+    let fatest = response
```
thedavidmeister, line 26: typo

orderbook-rs/src/interpreter/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+/// * `rpc_url` - Network RPC URL.
+///
+pub async fn get_disp(
+    deployer_npe2: Address,
+    rpc_url: String,
+) -> anyhow::Result<(Address, Address, Address)> {
+    let provider = match Provider::<Http>::try_from(rpc_url.clone()) {
+        Ok(provider) => provider,
+        Err(err) => {
+            error!("INVALID RPC URL");
```
thedavidmeister, line 25: probably if this is to be a crate we want an error enum rather than strings, generally anyhow is probably better for binaries than libs

this is because if we expect someone else to use this crate as a dependency, they may want to change the behaviour of their code based on what happens in this code, in which case an enum allows them to do something like `match error { RainOrderbookError::InvalidRPC => { ... } }`

orderbook-rs/src/interpreter/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+/// * `rpc_url` - Network RPC URL.
+pub async fn parse_rainstring(
+    parser_address: Address,
+    rainstring: String,
+    rpc_url: String,
+) -> anyhow::Result<(Bytes, Vec<U256>)> {
+    let provider = match Provider::<Http>::try_from(rpc_url.clone()) {
+        Ok(provider) => provider,
+        Err(err) => {
+            error!("INVALID RPC URL");
```
thedavidmeister, line 77: if you're using the same error in several locations as a return, it's more weight to the idea that you want an enum

orderbook-rs/src/registry.rs
```diff
@@ -0,0 +1,14 @@
+use alloy_sol_types::sol;
+use ethers::contract::abigen;
+
+sol!(IOrderBookV3, "../out/IOrderBookV3.sol/IOrderBookV3.json");
+
+abigen!(IParserV1, "../out/IParserV1.sol/IParserV1.json");
+abigen!(
+    IExpressionDeployerV3,
+    r#"[
```
thedavidmeister, line 9: magic strings should be constants if we can

thedavidmeister (changes requested):

orderbook-rs/src/gasoracle/mod.rs
Siddharth2207: Resolved.

orderbook-rs/src/gasoracle/mod.rs
Siddharth2207: Resolved

orderbook-rs/src/interpreter/mod.rs
Siddharth2207: Added error enum.

orderbook-rs/src/interpreter/mod.rs
Siddharth2207: Resolved.

orderbook-rs/src/registry.rs
Siddharth2207: Macros do not accept &str constants as parameters.

PR #57, opened 2024-01-09
tauri artifacts build

thedavidmeister:


mattyg:
I added the scaffolded tauri app, but we need these dependencies in nix to build it: https://tauri.app/v1/guides/getting-started/prerequisites#setting-up-linux

mattyg (approved):

PR #58, opened 2024-01-10
2024 01 09 tauri UI setup

mattyg:
Basic scaffolding of tauri app ui

- replace scaffolded svelte app with scaffolded sveltekit app
- add tailwind to sveltekit app
- add flowbite to sveltekit app
- implement basic empty pages with sidebar navigation


tauri-app/src/app.d.ts
```diff
@@ -0,0 +1,13 @@
+// See https://kit.svelte.dev/docs/types#app
+// for information about these interfaces
+declare global {
+	namespace App {
+		// interface Error {}
```
thedavidmeister, line 5: commented code

tauri-app/package.json
```diff
@@ -1,26 +1,46 @@
...
-    "svelte-check": "^3.4.6",
-    "tslib": "^2.6.0",
-    "typescript": "^5.0.2",
-    "vite": "^5.0.0",
-    "@tauri-apps/cli": "^1.5.8"
-  }
+	"name": "tauri-app",
+	"version": "0.0.1",
+	"private": true,
+	"scripts": {
```
thedavidmeister, line 5: why npm scripts?

thedavidmeister (changes requested):
we're using tabs for spaces?

everywhere else it's 2 spaces

mattyg:
> we're using tabs for spaces?
> 
> everywhere else it's 2 spaces

fixed & updated eslint config to 2 spaces

tauri-app/package.json
mattyg: generated by scaffolding and kept for reference but yes its redundant and unnecessary. removed.

tauri-app/src/app.d.ts
mattyg: removed

tauri-app/package.json
mattyg: reverted -- added npm scripts back in to ensure users are not running incorrect versions of node dependency commands by using `npx`

PR #61, opened 2024-01-11
2024 01 10 cli withdraw

mattyg:
Withdraw command mirroring patterns for deposit command in #54 

- minor refactoring of #54 
  -  move binding tests into separate dir
  -  reorder struct params to mirror abi
  -  extract common logic to execute a transaction call into a trait
  -  avoid glob reexport
- withdrawCall binding basic tests
- WidthdrawArgs struct
- withdraw cli command


crates/cli/src/execute.rs
```diff
@@ -0,0 +1,36 @@
...
+use alloy_ethers_typecast::transaction::ExecutableTransaction;
+use alloy_sol_types::SolCall;
+use anyhow::Result;
+use clap::Args;
+use clap::FromArgMatches;
+use clap::Parser;
+use rain_orderbook_common::transaction::TransactionArgs;
+
+#[derive(Parser)]
+pub struct ExecutableTransactionCall<T: FromArgMatches + Args> {
```
hardyjosh, line 11: @mattyg this is cool, only thing i'd say is this abstraction will probably break when we get to more complex commands like add-order, remove-order, take-orders etc where the cli args don't match the contract call args

mattyg: This just abstracts the logic for making a call with a ledger client and given transaction cargs -- conversion from cli args to contract call is still implemented manually for each cli command in the Execute trait.

mattyg:
A few issues are still outstanding:
- Deposit transaction fails (I believe because it doesn't call 'approve' on the ERC20 before calling deposit)
- Need to display better CLI output to user: 
  - "awaiting signature from ledger"
  - status of submitting transaction and/or loading spinner
  - successful transaction receipt hashes and block explorer link
- `ExecutableTransactionCall` abstraction may need to be removed / reworked to support making multiple transaction calls but will investigate further

PR #96, opened 2024-01-17
feat(cli): 'order detail' command

mattyg:
Basic 'order detail' cli command. Currently fetches way more data than we'll likely want in the detail view and displays it poorly.


thedavidmeister:
thisi was merged to main

PR #106, opened 2024-01-20
2024 01 19 gui vaults list page

mattyg:
Basic tauri page for Vaults list. Does not yet include filtering nor pagination.

image


mattyg:
i realized my subgraph queries need to be modified to distinguish by tokenVault -- please hold off review for now

thedavidmeister (approved):

PR #107, opened 2024-01-20
2024 01 19 gui vault detail page

mattyg:
Basic tauri page for Vault detail. Note that 'Deposit' and 'Withdraw' buttons don't do anything yet.

Also the deposits / withdraws tables are 'hoverable' in the ui, and shouldn't be -- since they aren't clickable.

image


mattyg:
i realized my subgraph queries need to be modified to distinguish by tokenVault -- please hold off review for now

thedavidmeister (approved):

PR #108, opened 2024-01-20
2024 01 19 gui deposit

mattyg:
Modals for 'Deposit' and 'Withdraw', opened by clicking buttons on Vault Detail page. Does not include actually making the deposit / withdrawal

image


tauri-app/src/lib/utils/number.ts
```diff
@@ -0,0 +1,3 @@
+export function isStringValidNumber(val: string) {
+  return /^\d+.?[\d]+$/.test(val);
```
thedavidmeister, line 2: can we get some test cases around this?

thedavidmeister (changes requested):

tauri-app/src/lib/utils/number.ts
mattyg: removed and replaced with [@imask/svelte](https://www.npmjs.com/package/@imask/svelte)

mattyg:
- replaced bespoke input mask implementation with 
- added "max" button to auto-fill the input with the max amount (currently only supported in Withdraw modal)
- added error message indicating that input is > max

tauri-app/src/lib/ModalVaultWithdraw.svelte
```diff
@@ -0,0 +1,87 @@
...
+      for="amount"
+      class="mb-2 w-full text-xl font-bold tracking-tight text-gray-900 dark:text-white"
+    >
+      Amount
+    </Label>
+    <InputTokenAmount
+      bind:value={amount}
+      bind:valueRaw={amountRaw}
+      symbol={vault.token_vaults ? vault.token_vaults[0].token.symbol : ''}
+      decimals={vault.token_vaults ? vault.token_vaults[0].token.decimals : 16}
```
thedavidmeister, line 66: 16?

thedavidmeister (changes requested):

tauri-app/src/lib/ModalVaultWithdraw.svelte
thedavidmeister: actually i think we should fallback to 0 with a clear error/warning that shows to the user that they have to put the raw absolute value in because we _can't_ infer the decimals for them

mattyg: it's resolved in PR #111 -- removed all the conditionals of vault.token_vaults

PR #127, opened 2024-01-23
fork draft

rouzwelt:


tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+use reqwest::Client;
+use revm::primitives::Bytes;
+use std::{collections::HashMap, sync::Mutex};
+
+/// static hashmap of fork evm instances, used for caching instances between runs
+pub static FORKS: Lazy<Mutex<HashMap<String, ForkedEvm>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+/// hashmap of cached error selectors    
+pub static SELECTORS: Lazy<Mutex<HashMap<Vec<u8>, Error>>> =
```
thedavidmeister, line 14: can this be a `[u8; 4]` rather than `Vec<u8>`?

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+#[tauri::command]
+pub async fn fork_call(
+    fork_url: &str,
+    fork_block_number: u64,
+    from_address: &[u8],
```
thedavidmeister, line 21: can these be address or `[u8; 20]` or something? maybe even something that implements https://docs.rs/serde_bytes/latest/serde_bytes/ ?

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+) -> Result<Result<Bytes, String>, String> {
+    let result = {
+        // lock static FORKS
+        let mut forks = FORKS.lock().unwrap();
+
+        // build key from fork url and block number
+        let key = fork_url.to_owned() + &fork_block_number.to_string();
+
+        // fork from the provided url, if it is cached, use it, if not create it, and cache it in FORKS
+        let forked_evm = if let Some(v) = forks.get_mut(&key) {
```
thedavidmeister, line 33: i think we can use entry here to handle the default empty case https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+            .map_err(|e| e.to_string())?
+    };
+
+    if result.reverted {
+        // decode result bytes to error selectors if it was a revert
+        Err(decode_error(&result.result).await?)
+    } else {
+        Ok(Ok(result.result))
+    }
+    // Ok(Ok(()))
```
thedavidmeister, line 58: commented code

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+    let selector_hash = alloy_primitives::hex::encode_prefixed(selector_hash_bytes);
+
+    // check if selector already is cached
+    {
+        let selectors = SELECTORS.lock().unwrap();
+        if let Some(error) = selectors.get(selector_hash_bytes) {
+            if let Ok(result) = error.abi_decode_input(args_data, false) {
+                return Ok(format!("{}: {:?}", error.name, result));
+            } else {
+                return Ok("unknown error".to_owned());
```
thedavidmeister, line 73: return an enum so the caller can handle the unknown error without string matching

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+    };
+
+    let url = "https://api.openchain.xyz/signature-database/v1/lookup";
```
thedavidmeister, line 78: magic string somewhere else like a const

tauri-app/src-tauri/src/main.rs
```diff
@@ -3,14 +3,16 @@
 
 mod commands;
 use commands::vault::{vault_detail, vaults_list};
-use commands::wallet::get_address_from_ledger;
+// use commands::wallet::get_address_from_ledger;
```
thedavidmeister, line 6: commented code

thedavidmeister (changes requested):

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: I used &[u8] to make it easier for using the underlying function, it will error if the u8 is not valid address length

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: this is meant to be shown on the problem panel, that's why it is string,

tauri-app/src-tauri/src/main.rs
rouzwelt: commented because it was part of the code, but was throwing errors, so I commented it, along with the wallet.rs

mattyg: It's throwing errors because you've pinned `alloy-ethers-typecast` to branch `2024-01-24-fix-ethers-dep` which is outdated

mattyg:
Just a note that we need to resolve the different requirements on alloy-ethers-typecast before merging this PR.

I need alloy-ethers-typecast to use ethers at [this hash](https://github.com/gakonst/ethers-rs/commit/439a0c7de0fb8569178d56933b000d8783805c2f) or newer (see https://github.com/rainlanguage/alloy-ethers-typecast/pull/10)

This PR uses alloy-ethers-typecast with ethers pinned to 0.2.11 (older than the above hash), due to requirements from https://github.com/rainlanguage/forker related to the ethers version used by foundry.

thedavidmeister (approved):

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,279 @@
...
+    if result.reverted {
+        // decode result bytes to error selectors if it was a revert
+        Ok(Err(decode_error(&result.result).await?))
+    } else {
+        Ok(Ok(result.result))
+    }
+}
+
+/// decodes an error returned from calling a contract by searching its selector in registry
+async fn decode_error(error_data: &[u8]) -> Result<DecodedErrorType, String> {
```
mattyg, line 83: mind if we move this function into the `common` crate? I'd like to use it in the cli too.

mattyg: or maybe it belongs in the WritableClient in alloy-ethers-typecast -- we can save that for a separate PR though

rouzwelt: yes sure

crates/common/src/error.rs
```diff
@@ -14,3 +26,77 @@ pub enum WritableTransactionExecuteError {
...
+    Unknown,
+    Known {
+        name: String,
+        args: Vec<String>,
+        sig: String,
+    },
+}
+
+/// decodes an error returned from calling a contract by searching its selector in registry
+pub async fn decode_error(error_data: &[u8]) -> Result<DecodedErrorType, String> {
```
thedavidmeister, line 41: please don't return strings as errors in crates

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,197 @@
...
+    fork_url: &str,
+    fork_block_number: u64,
+    from_address: ByteBuf,
+    to_address: ByteBuf,
+    calldata: ByteBuf,
+) -> Result<Result<Bytes, DecodedErrorType>, String> {
+    // build key from fork url and block number
+    let key = fork_url.to_owned() + &fork_block_number.to_string();
+
+    let is_cached = { FORKS.lock().unwrap().contains_key(&key) };
```
thedavidmeister, line 23: don't use unwrap in production code

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let result = if is_cached {
+        let mut forks = FORKS.lock().unwrap();
```
thedavidmeister, line 26: don't unwrap

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,197 @@
...
+        let forked_evm = forks.get_mut(&key).unwrap();
```
thedavidmeister, line 27: no unwrap

thedavidmeister (changes requested):

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: this unwrap is safe since it is checked before entring this arm of if statement

thedavidmeister (approved):

hardyjosh:
@mattyg would you prefer this to be more integrated into the way we're handling transaction calls? this looks like it assumes you're building the calldata on the front end, which isn't how we've been doing things so far.

in other words, in practice would this be exposed directly as tauri command?

mattyg:
> @mattyg would you prefer this to be more integrated into the way we're handling transaction calls? this looks like it assumes you're building the calldata on the front end, which isn't how we've been doing things so far.

Yeah I agree, it should just be in the `common` crate. I'll want to use it for both tauri and cli. I'll handle calling it via tauri commands as needed.

rouzwelt:
> > @mattyg would you prefer this to be more integrated into the way we're handling transaction calls? this looks like it assumes you're building the calldata on the front end, which isn't how we've been doing things so far.
> 
> Yeah I agree, i'd prefer it just be in the `common` crate. I'll want to use it for both tauri and cli. I'll handle calling it via tauri commands as needed.

applied

tauri-app/src-tauri/Cargo.toml
```diff
@@ -18,10 +18,15 @@ rain_orderbook_common = { path = "../../crates/common" }
 rain_orderbook_subgraph_client = { path = "../../crates/subgraph" }
 alloy-ethers-typecast = { git = "https://github.com/rainlanguage/alloy-ethers-typecast", rev = "9e830eda8a429fda60d1a87f480241c7678266bb" }
 alloy-primitives = "0.5.4"
-alloy-sol-types = "0.5.4"
 typeshare = "1.0.1"
+once_cell = "1.19.0"
```
mattyg, line 22: should be able to remove these imports from tauri now

crates/common/src/error.rs
```diff
@@ -14,3 +29,171 @@ pub enum WritableTransactionExecuteError {
...
+            }
+        }
+        Ok(AbiDecodedErrorType::Unknown)
+    } else {
+        Ok(AbiDecodedErrorType::Unknown)
+    }
+}
+
+#[derive(Debug)]
+pub enum AbiDecodeFailedErrors<'a> {
```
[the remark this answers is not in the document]
rouzwelt: thiserror doesn't allow for lifetimed errors variant on the enum as far as I checked, but anyways I've already have implemented these `From`s, so I guess it's fine?

Issue #79, opened 2024-01-14
Filling in gas fee values


mattyg:
Rather than expecting users to figure out the fees necessary to get their transactions validated in a reasonable timeframe, we should:
- Use a 3rd party api to determine the max_priority_fee_per_gas required for a slower, average, or faster transaction validation
- default to "average" priority fee
- In CLI:  add argument for transaction commands: `--priority-fee-speed` "slow", "average", "fast"
- In GUI: allow user to override priority-fee-speed for all transactions in a Settings page.

**todo**
- [ ] determine api for flare network gas estimation
- [ ] implement wrapper for gas estimation apis, potentially using ethers-rs gas oracle middleware
- [ ] modify cli to use gas estimation
- [ ] modify gui to use gas estimation

mattyg:
Just want to confirm this makes sense to you @thedavidmeister @hardyjosh - or if you have another suggestion.

hardyjosh:
@mattyg there's a gas oracle middleware for ethers - can we use that?

https://www.gakonst.com/ethers-rs/middleware/gas-oracle.html

mattyg:
> @mattyg there's a gas oracle middleware for ethers - can we use that?
> 
> https://www.gakonst.com/ethers-rs/middleware/gas-oracle.html

So the ethers gas oracle middleware just offers wrappers for a several different 3rd party apis. It looks like the only wrapper that implemented support for eip1559 estimation is blocknative. But unfortunately Blocknative only supports ethereum and polygon mainnets.

What chains do you need to support in the short term? And which do you aim to support longer-term? We may want to do some research into alternative 3rd party apis, choose one based on its pricing and chains support, and then implement the ethers gas oracle middleware for it ourselves. 

Another (more complex) option for an chain-agnostic gas estimation could be to make an `eth_feeHistory` rpc call and then calculate our own slow, average, fast fees for the prior block -- that would work fine when demand is relatively stable from one block to the next.

mattyg:
Flare's dev support did not know of any gas estimation api supporting the network:
![image](https://github.com/rainlanguage/rain.orderbook/assets/159270/b754ec45-ea9e-43cb-ae6a-b7a832717d6b)

mattyg:
Resolved by #167

PR #129, opened 2024-01-23
2024 01 23 gui handle deposit

mattyg:
Adds gui support for calling "deposit"

Currently the UX is weird, because the Deposit and Withdraw buttons are displayed on tokenVaults that aren't owned by your address. Will fix in a subsequent PR (https://github.com/rainlanguage/rain.orderbook/issues/135)

Currently the max_priority_fee_per_gas and max_fee_per_gas is hardcoded. Once I complete #79 I'll update the gui to call the same gas estimation rust code.

image
image
image

### DIFF
--- a/crates/common/src/deposit.rs
+++ b/crates/common/src/deposit.rs
@@ -1,5 +1,4 @@
-use alloy_primitives::{Address, U256};
-use anyhow::Result;
+use alloy_primitives::{hex::FromHexError, Address, U256};
 use rain_orderbook_bindings::{IOrderBookV3::depositCall, IERC20::approveCall};
 use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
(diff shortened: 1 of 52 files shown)


crates/common/src/deposit.rs
```diff
@@ -33,28 +34,8 @@ impl DepositArgs {
...
-        self.execute_approve(transaction_args.clone(), approve_transaction_status_changed)
-            .await?;
-        (approve_transaction_success)();
-        self.execute_deposit(transaction_args, deposit_transaction_status_changed)
-            .await?;
-
-        Ok(())
-    }
-
     /// Execute ERC20 approve call
```
mattyg, line 37: i felt like this helper function was unnecessary and just made it more confusing

thedavidmeister (approved):

PR #141, opened 2024-01-29
2024 01 28 gui remove order

mattyg:
Remove order in gui

- adds handler for remove order buttons
- button on Order Detail page opens generic confirmation dialog
- button on Order List page opens dialog with field to enter order id

image
image
image


hardyjosh:
@mattyg rather than a remove dialog that requires entering an order ID - perhaps that action could just be inline on each row in a context menu (i.e. ellipsis button)?

mattyg:
> @mattyg rather than a remove dialog that requires entering an order ID - perhaps that action could just be inline on each row in a context menu (i.e. ellipsis button)?

Yeah good call. I'll update.

mattyg:
Updated to:
- remove top-level "Remove" button from Order List page
- if order is active and order's owner is the wallet address specified in settings
   - display "Remove" button on Order Detail page, for orders owned by wallet address
   - display ellipsis button on Orders List page, for orders owned by wallet address
- if order is active and no wallet address specified in settings
  - display "Remove" button on Order Detail page, for every order
  - display ellipsis button on Orders List page, for every order
- ellipses button opens popover menu, clicking Remove button opens remove order confirmation modal

image
image
image

PR #171, opened 2024-02-03
codemirror ext - text editor

rouzwelt:


crates/common/src/fork.rs
```diff
@@ -54,6 +62,62 @@ pub async fn fork_call<'a>(
...
+    fork_block_number: u64,
+) -> Result<Bytes, String> {
+    // let rain_document = RainDocument::create(dotrain, None, None);
+    let deployer = AddOrderArgs::try_parse_frontmatter(front_matter)
+        .map_err(|e| e.to_string())?
+        .0;
+    // let entrypoints = ["calculate-order", "handle-io"];
+    // let rainlang_string = rain_document
+    //     .compose(&entrypoints)
+    //     .map_err(|e| e.to_string())?;
```
mattyg, line 79: commented code

crates/common/src/fork.rs
```diff
@@ -54,6 +62,62 @@ pub async fn fork_call<'a>(
     }
 }
 
+/// checks the front matter validity and returns the deployer address if check went ok
+pub async fn fork_parse_dotrain(
+    rainlang_string: &str,
+    front_matter: &str,
+    fork_url: &str,
+    fork_block_number: u64,
+) -> Result<Bytes, String> {
```
mattyg, line 71: can this return the actual error instead of a string?

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,13 @@
+use rain_orderbook_common::fork::fork_parse_dotrain;
+
+#[tauri::command]
+pub async fn fork_parse(
+    rainlang: &str,
+    front_matter: &str,
+    fork_url: &str,
+    fork_block_number: u64,
+) -> Result<(), String> {
```
mattyg, line 9: can this return a `CommandResult` from https://github.com/rainlanguage/rain.orderbook/blob/main/tauri-app/src-tauri/src/error.rs

tauri-app/src-tauri/src/commands/fork.rs
```diff
@@ -0,0 +1,13 @@
...
+    fork_parse_dotrain(rainlang, front_matter, fork_url, fork_block_number)
+        .await
+        .map(|_| ())
```
mattyg, line 12: do we want to return the error?

tauri-app/src/lib/texteditor/forkParse.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param forkUrl - frok url
+ * @param forkBlockNumber - fork block number
+ * @returns Resolves with empty array or with array of Problems if encountered an error
+ */
+export async function forkParseDotrain(dotrain: RainDocument, forkUrl: string, forkBlockNumber: number): Promise<Problem[]> {
+  let rainlang: string;
+  const frontMatter = dotrain.frontMatter;
+  try {
+    // set the hardcoded entrypoints
+    rainlang = await dotrain.compose(["calculate-io", "handle-io"]);
```
mattyg, line 17: I'd prefer to have the compose into the rust function in common.

mattyg (changes requested):
nice! works well. requested a few changes.

tauri-app/src/lib/texteditor/forkParse.ts
rouzwelt: well I didnt do that to not trigger another ound of dotrain parsing, the RainDocument instance passed here is already parsed, so there is no need to reparse it again in rust
also I dont really get the need for waht you are suggesting, this is only for text editor, I dont know how it being dont on rust side helps with anything

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: this command is specifically designed for text editor, which needs string as the error to show it on text editor

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: yes it can, I asked in the chat if you guys need it to return the expconfig (the bytes you asking for here), which everyone said no, so I skipped it

crates/common/src/fork.rs
rouzwelt: sure

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: this can return anything for the Ok variant, but text editor wont use it as we discussed, but feel free to change the Ok variant to wahtever you'd like

tauri-app/src-tauri/src/commands/fork.rs
rouzwelt: if you'd like you need to make a new wrapper command that returns the Error itself, but this one as is used in text editor should return the error as string so text editor can consume it

tauri-app/src/lib/texteditor/forkParse.ts
rouzwelt: this is specifically for text editor consumption, so if you'd need a similar function but the compose happening in rust, feel free to make that

rouzwelt: here the parsed RainDocument comes from the text editor, as the callback

rouzwelt: also this is living in rust, all RainDocument instances live in rust, not js, as they just interfaced through wasm bindings for js, so compose() here is actually calling on a rust command for composing

tauri-app/src-tauri/src/commands/fork.rs
mattyg: Yes that's the idea -- use the CommandError type to make a wrapper error. That error type can be serialized and displayed in the editor.

tauri-app/src/lib/texteditor/forkParse.ts
rouzwelt: the issue with what you are suggesting is that there will be another round of parsing on the dotrain string, since passing the RainDocument instance from js to tauri and then to rust will result in loosing the RainDocument instance, and it needs to built again with the original string,
or if you pass the RainDocument instance itself it will serialized/deserialized which again will result in a new instance of RainDocument which is not performant and not suitable for editor

rouzwelt: > > well I didnt do that to not trigger another round of dotrain parsing, the RainDocument instance passed here is already parsed, so there is no need to reparse it again in rust
> 
> You are calling fork_parse, after the compose succeeds. I'm saying do the compose and throw the error from the rust implementation in common. I'm going to use the same rust implementation for the cli, so I don't want duplicate code paths ideally.

if you want a command that composes the dotrain in rust side, I would say you need to make that one specifically, as it would be doing a different thing

mattyg: ah I see makes sense. I'll merge as-is and see if I can unify the code paths in a refactor.

mattyg (approved):

PR #179, opened 2024-02-06
WIP 2024 02 05 refactor fork

mattyg:
some refactoring to make the work in #171 and #127 a bit cleaner / more legible:
- split out abi decoding fn and types into separate file
- move fork call errors into fork.rs
- function rename

**todo**
- [ ] avoid deeply nested conditionals
- [ ] avoid double-wrapped return
- [ ] clarifying comments for constants
- [ ] pass around Address instead of byte arrays


mattyg:
moving this into another PR, some of the work has already been completed.

PR #193, opened 2024-02-08
2024 02 07 clear orders

mattyg:
- subgraph query to list order clears
- cli command `order-clear list`
- refactor all cli list commands to print csv formatted output to stdout, instead of writing to file with the `--csv` arg


crates/cli/src/commands/order_clear/list.rs
```diff
@@ -0,0 +1,91 @@
...
+            .to_subgraph_client()
+            .await?
+            .order_clears_list(pagination_args)
+            .await?;
+        let order_clears_flattened: Vec<OrderClearFlattened> = order_clears
+            .into_iter()
+            .map(|o| o.try_into())
+            .collect::<Result<Vec<OrderClearFlattened>, TryIntoFlattenedError>>()?;
+
+        if let Some(csv_file) = self.csv_file.clone() {
```
thedavidmeister, line 43: i think we want csv formatting but not file writing, this is a cli so it's easy for the caller to simply pipe the stdout to a file if they want that, but the inverse is more difficult

thedavidmeister (changes requested):

thedavidmeister (approved):

PR #243, opened 2024-02-15
2024 02 13 fuzz

hardyjosh:


crates/common/src/fuzz/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+};
+use strict_yaml_rust::StrictYamlLoader;
+use thiserror::Error;
+
+#[derive(Clone, Debug)]
+pub struct RunnerScenario {
+    pub name: String,
+    pub entrypoint: String,
+    pub runs: u64,
+    pub rebinds: Vec<String>,
```
thedavidmeister, line 22: can we not invent the word "rebinds" and just stick with "bind" which is used elsewhere?

crates/common/src/fuzz/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    fork::{ForkCallError, ForkedEvm, NewForkedEvm},
+    trace::RainEvalResult,
+};
+use strict_yaml_rust::StrictYamlLoader;
+use thiserror::Error;
+
+#[derive(Clone, Debug)]
+pub struct RunnerScenario {
+    pub name: String,
+    pub entrypoint: String,
```
thedavidmeister, line 20: now that we have quote rebinding, i'm not sure that we want/need to specify entrypoints

see https://github.com/rainlanguage/dotrain/issues/91 for a discussion on why/how

orderbook gui always uses the same entrypoints by convention, so introducing different entrypoints at the scenario level breaks the convention and causes the abstraction to leak

thedavidmeister (changes requested):

thedavidmeister (approved):

PR #306, opened 2024-02-17
2024 02 17 fix remove order vault

### LINKED ISSUE #302: Remove Order calls from gui sending incorrect vault id
likely an issue with hex / decimal conversion

```
my vault for the order is 0x11, but in the tx data for removeOrder its going as 17 

0000000000000000000000000000000000000000000000000000000000000017
should be 
0000000000000000000000000000000000000000000000000000000000000011
Thats why it was able to remove orders. @mattyjeee can you check this again once.
```

mattyg:
Resolves #302


crates/subgraph/src/types/order_detail_traits.rs
```diff
@@ -34,7 +34,7 @@ impl TryInto<OrderV2> for order_detail::Order {
                     Ok(IO {
                         token: v.token_vault.token.id.into_inner().parse::<Address>()?,
                         decimals: v.token_vault.token.decimals.try_into()?,
-                        vaultId: U256::from_str_radix(v.token_vault.vault_id.0.as_str(), 16)?,
+                        vaultId: U256::from_str_radix(v.token_vault.vault_id.0.as_str(), 10)?,
```
thedavidmeister, line 37: this doesn't seem right either, usually if something is `0x` prefixed it is treated as hex and otherwise it is treated as decimal

probably this parsing logic should be defined in a function that is reused across the app

thedavidmeister (changes requested):

crates/subgraph/src/types/order_detail_traits.rs
mattyg: I agree it seems weird. 

The vaultId string received from the subgraph is not prefixed by "0x" and is the base 10 value. I think in the replacement subgraph we'll want to remedy that and return a 0x-prefixed hex value. I agree this feels like a gotcha and will see where it makes sense to pull out / clarify the logic.

mattyg: Updated to use a helper util to convert base 10 string to u256 for clarity and added clarifying comment. I think at some point we may want to refactor to always convert the subgraph return types into our own types and pass those up to the cli and tauri.

PR #373, opened 2024-02-29
config ser deser

rouzwelt:


crates/settings/src/deployment.rs
```diff
@@ -0,0 +1,131 @@
...
+        let scenario = scenarios
+            .get(&self.scenario)
+            .ok_or(ParseDeploymentStringError::ScenarioNotFoundError(
+                self.scenario.clone(),
+            ))
+            .map(Arc::clone)?;
+
+        let order = orders
+            .get(&self.order)
+            .ok_or(ParseDeploymentStringError::ScenarioNotFoundError(
```
hardyjosh, line 41: @rouzwelt wrong error here

PR #380, opened 2024-03-01
fix config order in/out network match

### LINKED ISSUE #379: fix order input/ouput matching network

rouzwelt:
resolves #379

### DIFF
--- a/crates/settings/src/order.rs
+++ b/crates/settings/src/order.rs
@@ -29,6 +29,8 @@ pub enum ParseOrderStringError {
     TokenParseError(ParseTokenStringError),
     #[error("Network not found: {0}")]
     NetworkNotFoundError(String),
+    #[error("Network does not match")]
+    NetworkNotMatch,
 }
 
 impl OrderString {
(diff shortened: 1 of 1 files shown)


crates/settings/src/order.rs
```diff
@@ -70,31 +72,35 @@ impl OrderString {
...
-            .map(|output| {
-                tokens
-                    .get(&output)
-                    .ok_or(ParseOrderStringError::TokenParseError(
-                        ParseTokenStringError::NetworkNotFoundError(output.clone()),
-                    ))
-                    .map(Arc::clone)
-            })
-            .collect::<Result<Vec<_>, _>>()?;
+        let mut inputs = vec![];
```
mattyg, line 75: let's keep the implementation using iter map rather than for loops -- much clearer and easier to read.

mattyg (changes requested):
nice - but lets keep using iter + map rather than for loops for readability

crates/settings/src/order.rs
rouzwelt: cant really do this on an item, because need to check the network match on each input

PR #412, opened 2024-03-08
remove order's network

### LINKED ISSUE #401: remove network for order config in the yaml structure

rouzwelt:
resolves #401


thedavidmeister (changes requested):
i think this needs a corresponding update in the specs repo first

mattyg:
To me it feels less noisy / redundant to *only* set the network for the order, and then *not* have to set it for the deployer, input tokens and output tokens. I assume the they all have to be the same network, right?

rouzwelt:
> i think this needs a corresponding update in the specs repo first

I already made a comment on your Pr in specs repo

thedavidmeister (changes requested):
lets base this off the deployer's network rather than an arbitrary token

PR #424, opened 2024-03-14
performance forking

hardyjosh:


crates/common/src/fuzz/mod.rs
```diff
@@ -160,24 +169,35 @@ impl FuzzRunner {
...
+
+                let args = ForkEvalArgs {
+                    rainlang_string,
+                    source_index: 0,
+                    deployer: deployer.address,
+                    namespace: FullyQualifiedNamespace::default(),
+                    context: vec![],
+                    decode_errors: false,
+                };
+                fork_clone.fork_eval(args).await.unwrap()
```
thedavidmeister, line 191: these unwraps in these spawns will panic threads, can we handle results properly here?

thedavidmeister (changes requested):

crates/common/src/fuzz/mod.rs
hardyjosh: done

thedavidmeister (approved):

PR #428, opened 2024-03-15
walletconnect

rouzwelt:
**todo**
- [x] wallet connect qr code fails to load -- eventually displays "DECLINED" - I get a bunch of the following errors in console:
image
  -  [x] requires merging: https://github.com/rainlanguage/rainix/pull/49
- [x] wallet connect project id is hard-coded -- should read from env variable defined at build time
- [x] wallet connect modal should include all possible wallets that we have confirmed work
- [x] improve styling of buttons in forms to conect wallet connect, select a wallet option
image
image


rouzwelt:
improved buttons on select wallet modal
Screenshot 2024-03-17 at 12 57 37 AM

Screenshot 2024-03-17 at 12 57 43 AM

added all supported wallets:
Screenshot 2024-03-17 at 1 10 07 AM
Screenshot 2024-03-17 at 1 10 14 AM

mattyg:
nice work!

I confirmed connecting a wallet and calling all the contracts works for MetaMask mobile and Rainbow Wallet.

**A few issues I'd consider higher priority — we may want to fix them before merging:**
- WalletConnect transactions show no status notices, error messages or success messages (ideally this would use the same UX as current transactions but that will require a bigger refactor).
- Clicking the Remove button on the Order Detail page immediately initiates the transaction via ledger.
- If a ledger wallet address is saved in settings, action buttons & context menus are only be shown on Orders/Vaults owned by that wallet. This no longer makes sense with a multi-wallet system and should probably just be removed for the short-term.

**Other issues are low priority I think we can save for a subsequent PR**
- In all wallet selection forms: after clicking "proceed" and then "back", the modal fields are blank -- would be good to keep their contents
- Manually test all the most popular walletconnect wallets, verify connecting & calling works, remove all others
- refactor to use a common component for choosing a wallet via any connection method, rather than duplicating code in multiple components
- consider opening a separate modal to choose a deposit, rather than altering the contents of a model
- cleanup ui for choosing a connection method & wallet

rouzwelt:
> * WalletConnect transactions show no status notices, error messages or success messages (ideally this would use the same UX as current transactions but that will require a bigger refactor).

the status is shown in your wallet, thats how walletconnect works.

> * Clicking the Remove button on the Order Detail page immediately initiates the transaction via ledger.

need to investigate this
 
> **Other issues are low priority I think we can save for a subsequent PR**
> 
> * In all wallet selection forms: after clicking "proceed" and then "back", the modal fields are blank -- would be good to keep their contents

this is because mask component thing used on those fields

> * Manually test all the most popular walletconnect wallets, verify connecting & calling works, remove all others

they all connect under wallet connect, it is just shown with a different image on the modal

rouzwelt:
fixed high priority issues

mattyg:
Running on macOS 11 I get a white screen with the following console error:
image

Need to investigate further -- maybe a newly added ui library requires JS features that safari 14 doesn't support?

tauri-app/src/lib/components/ModalVaultDeposit.svelte
```diff
@@ -3,91 +3,139 @@
...
+      // eslint-disable-next-line no-console
+      console.log(e);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (typeof e === "object" && (e as any)?.reason) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        toasts.error(`Transaction failed, reason: ${(e as any).reason}`);
+      }
+      else if (typeof e === "string") toasts.error(e);
+      else if (e instanceof Error) toasts.error(e.message);
+      else toasts.error("Transaction failed!");
```
mattyg, line 64: If the logic to handle and error and convert it into a string is shared between multiple components, can we pull it out into a helper util?

tauri-app/src/lib/components/ModalExecute.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="ts">
```
mattyg, line 1: I think ideally this component would behave more like the other generic InputX components, where they expose a single prop 'value' and provide a generic input to choose a wallet of any type. It would require some more refactoring of how we're handling wallets in the store though. I'll make a separate issue for it. For now I think this is good.

tauri-app/src-tauri/src/commands/order.rs
```diff
@@ -115,3 +116,52 @@ pub async fn order_remove(
...
+
+#[tauri::command]
+pub async fn order_add_calldata(
+    app_handle: AppHandle,
+    dotrain: String,
+    deployment: Deployment,
+    transaction_args: TransactionArgs,
+) -> CommandResult<Bytes> {
+    let add_order_args = AddOrderArgs::new_from_deployment(dotrain, deployment).await?;
+    let calldata = add_order_args.get_add_order_calldata(transaction_args)
```
mattyg, line 128: I think its a bit of a shame that we can't use the WritableTransaction from alloy-ethers-typecast to make the actual contract call -- as we loose the really nice transaction status notice implementation. That said it seems like it would be very complex to implement a walletconnect SingerMiddleware for ethers-rs, and is probably not worth it for now.

tauri-app/src/lib/components/ModalVaultDepositGeneric.svelte
```diff
@@ -1,69 +1,117 @@
...
 
-  <div>
-    <h5 class="mb-2 w-full text-xl font-bold tracking-tight text-gray-900 dark:text-white">
-      Token
-    </h5>
-    <InputToken bind:address={tokenAddress} bind:decimals={tokenDecimals} />
-  </div>
+      const depositCalldata = await vaultDepositCalldata(vaultId, tokenAddress, amount) as Uint8Array;
+      const depositTx = await ethersExecute(depositCalldata, $orderbookAddress);
+      toasts.success("Transaction sent successfully!");
```
mattyg, line 55: It would be great to use the same styling for transaction status notices for both ledger and wallet connect transactions, but it would require a some work to wrap ethers transactions, determine their progress state, and pass that up to a transaction status notice.

I feel like for now this is good and we can do that in a subsequent PR.

tauri-app/src/lib/components/ModalVaultDeposit.svelte
mattyg: i implemented this

tauri-app/package.json
```diff
@@ -50,8 +50,10 @@
     "@observablehq/plot": "^0.6.13",
     "@square/svelte-store": "^1.0.18",
     "@tauri-apps/api": "^1.5.3",
+    "@web3modal/ethers5": "^4.1.1",
     "codemirror-rainlang": "^3.0.3",
     "dayjs": "^1.11.10",
+    "ethers": "^5.7.2",
```
mattyg, line 56: Its too bad we couldn't get wagmi / viem working -- viem is sooo nice. But I'm happy to have something working for now.

mattyg (approved):

rouzwelt:
@mattyg why do you want to replace ethers with wagmi?

tauri-app/package.json
rouzwelt: i dont think that really matters, the main issue was wagmi part of it, and i dont think just because viem is so nice we should automatically go for wagmi because it is based on viem?
i worked on wagmi for a few days, before switching to ethers, and personally ethers implementation is much simpler/cleaner compared to wagmi.
but after all they both do the same job and offer pretty much same funtionalities

mattyg:
> @mattyg why do you want to replace ethers with wagmi?

Just a way more pleasant, ergonomic dev experience with wagmi & viem.

tauri-app/src/lib/components/ModalExecute.svelte
mattyg: #481

tauri-app/src/lib/components/ModalVaultDepositGeneric.svelte
mattyg: #482

PR #432, opened 2024-03-16
2024 03 15 deposit allowance check

### LINKED ISSUE #403: Check approve allowance before calling approve
If approve allowance is already enough, skip straight to deposit

mattyg:
Resolves #403 

- Check allowance before calling approve
- Only request approval for allowance - amount

### DIFF
--- a/crates/common/src/deposit.rs
+++ b/crates/common/src/deposit.rs
@@ -1,8 +1,36 @@
-use crate::transaction::{TransactionArgs, WritableTransactionExecuteError};
-use alloy_ethers_typecast::transaction::{WriteTransaction, WriteTransactionStatus};
+use crate::transaction::{TransactionArgs, TransactionArgsError, WritableTransactionExecuteError};
+use alloy_ethers_typecast::{
+    ethers_address_to_alloy,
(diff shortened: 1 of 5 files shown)


crates/common/src/deposit.rs
```diff
@@ -22,30 +50,56 @@ impl From<DepositArgs> for depositCall {
...
-            amount: self.amount,
-        }
+    /// Execute read IERC20 allowance call
+    pub async fn read_allowance(
+        &self,
+        owner: Address,
+        transaction_args: TransactionArgs,
+    ) -> Result<U256, DepositError> {
+        let readable_client =
+            ReadableClient::new_from_url(transaction_args.rpc_url.clone()).unwrap();
```
thedavidmeister, line 60: this will panic the app if you unwrap

thedavidmeister (changes requested):

crates/common/src/deposit.rs
mattyg: whoops - good catch

thedavidmeister (approved):

PR #478, opened 2024-03-21
fix(tauri/ui): missing types

mattyg:
vite builds are failing locally for me with the following message:

```
Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point for implicit type library 'node'
```

This PR resolves it -- although builds are not failing on CI, so I'm not clear on my local issue. Leaving this open for now.


mattyg:
superseded

PR #617, opened 2024-05-04
2024 05 04 matchstic

### LINKED ISSUE #616: add matchstick

thedavidmeister:
fixes #616


subgraph/networks.json
```diff
@@ -0,0 +1,8 @@
+{
+    "matic": {
+        "metaboard0": {
```
thedavidmeister, line 3: this was taken verbatim from the metaboard setup but doesn't make any sense

will need to be replaced with an orderbook address when one exists

subgraph/package.json
```diff
@@ -5,6 +5,8 @@
     "author": "rainlanguage",
     "license": "CAL",
     "dependencies": {
-      "@graphprotocol/graph-ts": "=0.32.0"
+        "@graphprotocol/graph-cli": "^0.56.0",
```
thedavidmeister, line 8: this dep ideally wouldn't be here due to graph cli existing in nix, but it has transitive deps that matchstick silently relies on (will error when matchstick tries to run)

PR #633, opened 2024-05-08
Adding Deposit, Transaction entities, Event interface

hardyjosh:
## Motivation

We're build a new subgraph and need the first entities and tests.

## Solution

Added Deposit. Also added a reusable interface for Events and a reusable entity for Transaction.

Will replace the addresses like `sender` with an `Account` entity later.

## Checks


subgraph/src/deposit.ts
```diff
@@ -1 +1,19 @@
...
+import { Deposit as DepositEntity } from "../generated/schema";
+import { createTransactionEntity } from "./transaction";
+
+export function handleDeposit(event: Deposit): void {
+  createDepositEntity(event);
+}
+
+export function createDepositEntity(event: Deposit): void {
+  let deposit = new DepositEntity(
+    event.transaction.hash.toHex() + "-" + event.logIndex.toString()
```
thedavidmeister, line 11: this can be bytes i think

thedavidmeister (changes requested):

subgraph/src/deposit.ts
hardyjosh: fixed now

PR #641, opened 2024-05-14
filter elided bindings on lsp

### LINKED ISSUE #640: Squash elided bindings errors in extension
The Raindex app currently shows errors for elided bindings that may actually be set in one or more scenarios. This is distracting, can we have a way to squash these errors in the codemirror extension? This might be best to do via a flag, as the error would still be useful when composing a specific scenario.

image

rouzwelt:
resolves #640


hardyjosh:
This is a workaround that doesn't address the root cause of the issue, those are further discussed in #643.

Closing this PR for now.

PR #646, opened 2024-05-14
random vault id

### LINKED ISSUE #605: Allow for unset vault id in order front matter yaml and generate a random one

rouzwelt:
## Motivation
resolves #605 

## Solution
use `rand` lib to generate random `U256` for optional vault ids in config

## Checks

### DIFF
--- a/crates/common/src/add_order.rs
+++ b/crates/common/src/add_order.rs
@@ -65,12 +65,13 @@ impl AddOrderArgs {
         dotrain: String,
         deployment: Deployment,
     ) -> Result<AddOrderArgs, AddOrderArgsError> {
+        let random_vault_id = random_u256();
         let mut inputs = vec![];
         for input in &deployment.order.inputs {
             if let Some(decimals) = input.token.decimals {
                 inputs.push(IO {
                     token: input.token.address,
-                    vaultId: input.vault_id.unwrap_or(random_u256()),
+                    vaultId: input.vault_id.unwrap_or(random_vault_id),
                     decimals,
                 });
             } else {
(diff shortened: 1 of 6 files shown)


crates/common/src/add_order.rs
```diff
@@ -70,7 +70,7 @@ impl AddOrderArgs {
             if let Some(decimals) = input.token.decimals {
                 inputs.push(IO {
                     token: input.token.address,
-                    vaultId: input.vault_id,
+                    vaultId: input.vault_id.unwrap_or(random_u256()),
```
thedavidmeister, line 73: isn't this going to produce a different vault id for every input and output?

that seems like it's obviously not what people want/expect

thedavidmeister (changes requested):

hardyjosh (changes requested):
One of the big use cases for Raindex is the ability to rotate your input/output tokens by setting a token as both the input and output (e.g. @thedavidmeister's flipper dipper strat).

This won't work if the input and output have different vault ids. We need to pick a single random vault id and apply it to all `None` vault ids in the order.

**Also - this PR is missing tests. There's no test that shows when a vault id is not set in an order, a random one will be created.**

crates/common/src/add_order.rs
```diff
@@ -232,10 +233,31 @@ impl AddOrderArgs {
     }
 }
 
+pub fn random_u256() -> U256 {
```
thedavidmeister, line 236: Why do we need this function at all? instead of just calling `rand::random()` directly?

if we're intending for this function to be used across all our crates consistently this is the wrong place for it to live, if we aren't then i would say to delete it

thedavidmeister (changes requested):

thedavidmeister (approved):

hardyjosh (approved):

Issue #674, opened 2024-06-04
Add backtesting to raindex


hardyjosh:
We can already do montecarlo fuzzing for strategies. Users specify a number of `runs` in a scenario, and any elided bindings in the rainlang that are not given a value in the scenario are replaced with a random 32-bytes for each run.

The spec for that is here: https://github.com/rainlanguage/specs/blob/main/ob-yaml.md

The stacks are flattened out and users can build charts, for example creating a scatterplot where the x and y for each point are specific stack indexes from indiviual runs.

Currently this happens on a local fork of the latest block. What we'd like to add is the ability to specify a range of historical blocks.

- add a new field to [Scenario](https://github.com/rainlanguage/rain.orderbook/blob/main/crates/settings/src/scenario.rs), `blocks`, where a user can specify a block range, e.g. `[5000, 10000]`
- create a new outer loop around the fuzz runs that rolls the fork to each block in the range

We actually don't need to add anything special after this, if someone wants to chart block against some other stack item, they can put the `block-number()` on the stack in their rainlang.

hardyjosh:
this was implemented in the end

PR #681, opened 2024-06-15
2024 06 14 backtesting

hardyjosh:
## Motivation

Fix for https://github.com/rainlanguage/rain.orderbook/issues/674

## Solution

Implemented simple "backtesting", ie being able to run simulations across a range of blocks. Most of the work was create the deserializer for the syntax we wanted to use, and converting the range syntax into a concrete `Vec` of `BlockNumbers`. 

After that adding to the actual simulation loop was straightforward, except that Foundry's `roll_fork` doesn't actually update the env blocknumber so the rainlang word `block-number()` wasn't working.

To fix that I had to update the env block number every time we roll here: https://github.com/rainlanguage/rain.interpreter/pull/251

That PR needs to be merged first before merging this PR. 

## Checks


crates/common/src/fuzz/mod.rs
```diff
@@ -211,55 +229,59 @@ impl FuzzRunner {
...
+                let scenario_bindings = scenario_bindings.clone();
+                let dotrain = Arc::clone(&dotrain);
+
+                let mut final_bindings: Vec<Rebind> = vec![];
+
+                // For each scenario.fuzz_binds, add a random value
+                for elided_binding in elided_binding_keys.as_slice() {
+                    let mut val: [u8; 32] = [0; 32];
+                    self.rng.fill_bytes(&mut val);
+                    let hex = format!("0x{}", alloy_primitives::hex::encode(val));
```
thedavidmeister, line 252: does this ensure even number of hex nibbles on the output? i.e. left pad with 0 til even number of chars

crates/settings/src/blocks.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+impl Default for Blocks {
+    fn default() -> Self {
+        Blocks::RangeWithInterval {
+            range: BlockRange {
+                start: Block::Genesis,
+                end: Block::Latest,
+            },
+            interval: 1,
```
thedavidmeister, line 176: hmm, a safer default would be start/end both at latest, this will expand out to millions of iterations probably

thedavidmeister (changes requested):

crates/settings/src/blocks.rs
hardyjosh: ah yeah fair, i put this in for other tests where we need a default, but actually it's an `Option` in Scenario anyway, so I can delete this.

crates/common/src/fuzz/mod.rs
hardyjosh: i've added padding to the format string

hardyjosh:
@thedavidmeister assigning this to you so you can resolve the contract import issues

PR #701, opened 2024-06-25
ERC20 Entity for Subgraph

hardyjosh:
## Motivation

Pinging an RPC many times to pull token details doesn't really make sense when the name/symbol/decimals can be grabbed once, the first time the token appears in the subgraph.

## Solution

Added an ERC20 entity to Vault. Removed the token field from all other entities, as the Vault is always the logical reference from anywhere else.

## Checks

### DIFF
--- a/subgraph/schema.graphql
+++ b/subgraph/schema.graphql
@@ -1,7 +1,7 @@
 type Vault @entity {
   id: Bytes!
   "The token that this vault is for"
-  token: Bytes!
+  token: ERC20!
   "The owner of this vault"
   owner: Bytes!
   "The vaultId of this vault"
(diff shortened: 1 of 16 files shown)


subgraph/schema.graphql
```diff
@@ -199,4 +195,16 @@ interface Event {
   transaction: Transaction!
   "msg.sender for the event"
   sender: Bytes!
+}
+
+type ERC20 @entity(immutable: true) {
```
thedavidmeister, line 200: if someone was to deploy an ERC20 that changed these values, and then traded it, would that brick the sg due to this `immutable`?

thedavidmeister (changes requested):

hardyjosh:
No, because the sg doesn’t update it. It only creates it the first time.

subgraph/schema.graphql
hardyjosh: No, because the sg doesn’t update it. It only creates it the first time.

thedavidmeister (approved):

PR #727, opened 2024-07-21
Tanstack app table component

hardyjosh:
## Motivation

We want to move to using Tanstack Query, so that we get caching, invalidation, etc and out of the box.

## Solution

image

This is a rewrite of our existing AppTable component, but for Tanstack. Pagination is replaced with infinite queries.

## Checks


thedavidmeister (changes requested):
can i get a screenshot on this?

PR #744, opened 2024-07-29
fixes for order id matching when using a bigint orderhash

hardyjosh:
## Motivation

We had issues with Meta, AddOrder and RemoveOrder entities not being linked to the correct orders, because of the way the Order ids were being calculated. In particular, it looks like the bytes for a Uint256 are reversed compared to the same bytes being emitted in the AddOrder event. I was also inexplicably getting leading zeroes.

## Solution

Fixed both bugs.

## Checks


hardyjosh:
Closing this as it won't be relevant once the event emitted after an order is added is updated to where the subject is bytes.

PR #794, opened 2024-08-26
Subgraph schema validator

### LINKED ISSUE #789: Logic in rain.orderbook common crate for comparing a remote schema against the version in the app/repo and a wrapper CLI command
This issue needs a bit of planning, we can probably use cynic as a crate to achieve this

see https://cynic-rs.dev/schemas/introspection

We don't want to use the cynic cli, we want our own cli for orderbook, e.g.

```
orderbook subgraph validate https://api.goldsky.com/api/public/project_clv14x04y9kzi01saerx7bxpg/subgraphs/ob-flare-0xb06202aA/1.0.0/gn
```

This would simply:
- introspect the remote URL to get the schema
- compare it against the schema that is hardcoded into the subgraph crate `https://github.com/rainlanguage/rain.orderbook/blob/main/crates/subgraph/schema/orderbook.graphql`

If it doesn't match it's an error. If we use the same source code that cynic uses this should line up perfectly if it's compatible, because the hardcoded .graphql file came from the cynic cli.

rouzwelt:
## Motivation
resolves #789 

## Checks


crates/subgraph/src/validate.rs
```diff
@@ -0,0 +1,76 @@
...
+    Ok(main_schema == schema)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_validate_happy_true() {
+        let subgraph_url = "https://api.goldsky.com/api/public/project_clv14x04y9kzi01saerx7bxpg/subgraphs/ob4-test/0.7/gn";
```
hardyjosh, line 58: can we do a mock instead? these apis won't necessarily be up forever

PR #887, opened 2024-09-25
Deployment debug tab on add order

### LINKED ISSUE #867: Replace the debug panel on add order with something that works just like quotes on the live order pages

findolor:
Closes #867

## Motivation

See issue: #867

## Solution

- Decoupled the debug logic from run scenario button. It now fetches the correct data after each keystroke using a debounced function
- Added a visual component similar to order quotes
Screenshot 2024-09-25 at 11 46 18
Screenshot 2024-09-25 at 11 48 05

## Checks


hardyjosh:
So @findolor this looks great on the happy path but i found a couple of issues:
1. Simple one - if there's only one thing on the calculate-io stack, the component breaks. I assume because it's looking for [0] and [1] and only [0] is there. If this is the case it should just say there isn't enough on the stack to get both a max amount and ratio.
2. Another simple one - if ratio is 0 i get this error 
image

3. For these tokens I'm getting "UNKNOWN/UNKNOWN" as both pairs?
```
tokens:
  polygon-usdt:
    network: polygon
    address: 0xc2132D05D31c914a87C6611C10748AEb04B58e8F
    decimals: 6
  polygon-ioen:
    network: polygon
    address: 0xd0e9c8f5fae381459cf07ec506c1d2896e8b5df6
    decimals: 18

orders:
  polygon:
    orderbook: polygon
    inputs:
      - token: polygon-ioen
      - token: polygon-usdt
    outputs:
      - token: polygon-ioen
      - token: polygon-usdt
```

4. If there's any error at all with the rainlang or yaml the spinner spins indefinitely and no error is shown. We need to show the error, here's a super rough example
image

5. The thing is, only some pairs might have an error. For example this one should only show an error for one of the pairs, but an error for any pair means the whole call fails. The other thing is we need to show partial traces, so if there is an error we still need the trace up until the error, as we do with quotes.
image

hardyjosh:
If I have a really long value (in this case it's`max-value()`), then the hex version gets pushed off the side and i can't scroll.

might need an `overflow-x-scroll` wrapper on the table.

hardyjosh:
something is weird where the older error isn't being removed when new information comes through

image

hardyjosh:
still getting UNKNOWN/UNKNOWN for the pairs

image

hardyjosh (changes requested):
See comments for manual testing feedback

hardyjosh:
@findolor we're still missing partial stacks - if you get a revert for some quote you can still load up whatever stack was produced before the revert. You can see how i implemented that, i thought it would be the same here.

So for example, in the below image i'd expect to be able to load up the stack and see 14, because that was produced before the revert.

image

hardyjosh:
@findolor can we have the block paused by default? when the block is constantly updating it has to get a new fork every time which makes it quite slow to produce the debug. but if it's paused it's much faster (same fork every time so it's all local).

hardyjosh:
@findolor also - i know we didn't want to upgrade anything to do with charting in this PR, but let's at least rename the "Run all scenarios" button to "Generate charts"

hardyjosh:
replaced by #1599

PR #890, opened 2024-09-25
Remove apply settings button

### LINKED ISSUE #452: Remove 'apply settings' button

findolor:
Closes #452

## Motivation

See issue: #452

## Solution

- Removed apply settings button
- Updated settings store variable to only hold the settings object
- Parse the input text in settings page and show update failed toast on error

Screenshot 2024-09-25 at 18 18 50

## Checks


hardyjosh (commented):
@findolor definitely better than before! But I have some thoughts...

Instead of a toast on error and nothing on success, i think it would be better to have a persistent status at the top, instead of the "settings" title. it could be an Alert component (like the top bar that says "looking for some settings to get started?")

- Grey (with spinner) - "Checking settings". show after each keystroke until the debounce is resolved
- Green (with tick) - "Settings applied"
- Red (with cross) - "Error applying settings -see below"

Also i think the text box should fill the screen and be scrollable inside, so that the top part of the screen with the status, save as and load buttons remains fixed.

findolor:
@hardyjosh what do you think about this look?

Screenshot 2024-09-26 at 19 07 19
Screenshot 2024-09-26 at 19 07 39
Screenshot 2024-09-26 at 19 07 46
Screenshot 2024-09-26 at 19 07 54

The code section can now grow in size based on screen's height. It looks good on both small and big screens.

hardyjosh:
@findolor looks really good - I would just say that we can probably replace the "settings" heading and put this new alert in its place. So the alert and buttons would make one row. The user knows from the breadcrumb at the top that they're on the settings page (if it isn't already obvious), so I don't think we need the additional heading.

findolor:
Screenshot 2024-09-27 at 13 32 27
Screenshot 2024-09-27 at 13 32 37
Screenshot 2024-09-27 at 13 32 47

hardyjosh:
@findolor i think we just need the file path added back in right? then this should be good to go!

findolor:
@hardyjosh file path is now underneath the buttons

Screenshot 2024-09-30 at 09 08 35

hardyjosh (approved):

PR #912, opened 2024-10-01
order dotrain meta

### LINKED ISSUE #888: Add the dotrain for a deployment to the meta for an order, as well as the composed rainlang

### LINKED ISSUE #889: Change the "rainlang" tab on the order page to "source" and add subtabs for the rainlang and the dotrain

rouzwelt:
## Motivation
resolves #888 
resolves #889 

## Solution
- add `gui` field to settings as `any` type ie `serde_json::Value`
- compress metas before emitting
- emit filtered dotrain meta when adding order and decode it back from sg order query meta field
- add dotrain ui tab for order page

## Checks

### DIFF
--- a/crates/cli/src/commands/order/filter.rs
+++ b/crates/cli/src/commands/order/filter.rs
@@ -28,6 +28,10 @@ pub struct Filter {
     /// Print the result on console (send result to std out)
     #[arg(long, action = ArgAction::SetTrue)]
     pub stdout: bool,
+
+    /// Include 'gui' field in final result
+    #[arg(long, action = ArgAction::SetTrue)]
+    pub include_gui: bool,
 }
 
 impl Execute for Filter {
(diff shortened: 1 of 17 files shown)


crates/cli/src/commands/order/filter.rs
```diff
@@ -28,6 +28,10 @@ pub struct Filter {
     /// Print the result on console (send result to std out)
     #[arg(long, action = ArgAction::SetTrue)]
     pub stdout: bool,
+
+    /// Include 'gui' field in final result
+    #[arg(long, action = ArgAction::SetTrue)]
+    pub include_gui: bool,
```
thedavidmeister, line 34: why wouldn't we include gui?

thedavidmeister (changes requested):

dcatki:
@rouzwelt this also

thedavidmeister:
the way this is is not useable for several reasons
- gui is way too big to include in full
- setting a bool as the author to include the gui or not doesn't make sense
- yaml implementation is being refactored in parallel

@rouzwelt @hardyjosh feel free to reopen if you disagree but i think we'll just have to revisit later

PR #952, opened 2024-10-21
Handle negative values and division operation

findolor:
## Motivation

See issue: #944

## Solution

- Multiply output value with -1
- Use `BigInt` values to correctly handle division

## Checks

### DIFF
--- a/tauri-app/src/lib/components/tables/OrderTradesListTable.svelte
+++ b/tauri-app/src/lib/components/tables/OrderTradesListTable.svelte
@@ -83,41 +83,31 @@
     </TableBodyCell>
     <TableBodyCell tdClass="break-all py-2">
       {formatUnits(
-        BigInt(item.outputVaultBalanceChange.amount),
+        BigInt(item.outputVaultBalanceChange.amount) * BigInt(-1),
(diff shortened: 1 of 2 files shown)


thedavidmeister (changes requested):
screenshots and tests

findolor:
All the trades render without negative values now:

Screenshot 2024-10-23 at 07 37 29

tauri-app/src/lib/components/tables/OrderTradesListTable.svelte
```diff
@@ -83,41 +83,31 @@
...
-          ),
-        ) /
-          Number(
-            formatUnits(
-              BigInt(item.outputVaultBalanceChange.amount),
-              Number(item.outputVaultBalanceChange.vault.token.decimals ?? 0),
-            ),
-          ),
+      {formatUnits(
+        (BigInt(10 ** 18) *
```
thedavidmeister, line 93: this should probably be abstracted in the same way that it is onchain

e.g. `a.mul(b)` is like `(a * b) / ONE` and `a.div(b)` is `(a * ONE) / b`

nobody writes solidity like this, for a reason :)

thedavidmeister (changes requested):

Issue #964, opened 2024-10-25
sketch for gui struct in rust


hardyjosh:
# Interface Definition for DotrainOrderGui

TODO:
- think about how we can modify tokens and when in the process this happens, especially if we have contextual replacements in bindings etc

# Core Interfaces

## DotrainOrder
```
// The underlying order implementation that handles RainLang composition and deployment
(As per attached implementation)
```

## Deposit 
```rust
// Configuration for what token deposits are allowed for a deployment
struct Deposit {
   token: String,      // Token identifier matching the YAML config
   min: f64,          // Minimum amount that can be deposited
   presets: Vec<f64>  // Suggested/common deposit amounts to show in UI
}
```

## Preset
```rust
// A preset value for a field with a human readable name
struct Preset {
   name: String,     // Human readable name for the preset (e.g. "Every hour (3600)")
   value: FieldValue // The actual value for this preset
}
```

## FieldDefinition
```rust
// Definition of a configurable field in a deployment
struct FieldDefinition {
   id: String,         // Unique identifier (hash of the field definition)
   name: String,       // Human readable field name
   description: String, // Detailed field description/help text
   validation: String,  // JSON Schema string for field validation
   presets: Vec<Preset> // Suggested/common values for this field
}
```

## Deployment
```rust
// A complete deployment configuration from the YAML
struct Deployment {
   id: String,           // Deployment identifier
   name: String,         // Human readable name
   description: String,  // Detailed deployment description
   deposits: Vec<Deposit>, // Allowed token deposits
   fields: Vec<FieldDefinition> // Configurable fields
}
```

## GUI
```rust
// Main state container for order configuration
struct DotrainOrderGui {
   dotrain_order: DotrainOrder,  // Reference to underlying order
   deployment: Deployment,        // Selected deployment configuration
   field_values: HashMap<String, FieldValue>, // Current values for fields
   deposits: Vec<TokenDeposit>,  // Current token deposits
}
```

### Constructor
```rust
// Creates new GUI for configuring a specific deployment
DotrainOrderGui::new(dotrain_order: DotrainOrder, deployment_id: String) -> Result<Self, GuiError>
```

### Field Operations
```rust
// Save single field value with validation
save_field(field_id: String, value: FieldValue) -> Result<(), ValidationError>

// Save multiple field values at once with validation
save_fields(field_values: Vec<(String, FieldValue)>) -> Result<(), Vec<ValidationError>>

// Get current value for a field
[...]

PR #1003, opened 2024-11-15
Adding order related operations to gui bindings

findolor:
## Motivation

See issue: #964

## Solution

- Added various operations for order related actions

## Checks


crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,293 @@
...
+            .outputs
+            .iter()
+            .filter(|output| !all_vaults.contains(*output))
+            .cloned()
+            .collect();
+        all_vaults.extend(filtered_outputs);
+        all_vaults
+    }
+
+    fn get_deposits_as_map(&self) -> HashMap<Address, String> {
```
hardyjosh, line 66: deposits should be U256

crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,293 @@
...
+        Ok(decimals)
+    }
+
+    async fn get_deposit_args(
+        &self,
+        rpc_url: &str,
+        deposits_map: &HashMap<Address, String>,
+        order_io: &OrderIO,
+    ) -> Result<DepositArgs, GuiError> {
+        let decimals = if let Some(decimals) = order_io.token.decimals {
```
hardyjosh, line 92: it's not really safe to use the decimals in the yaml - i think we should always check onchain

what would be good is to store all the token info (name/symbol/decimals) in memory, and allow it to be prefetched. then here we would either use the info already in memory, or fetch if we have to

crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,293 @@
...
+            allowance,
+        })
+    }
+
+    /// Check allowances for all inputs and outputs of the order
+    ///
+    /// Returns a vector of [`TokenAllowance`] objects
+    #[wasm_bindgen(js_name = "checkAllowances")]
+    pub async fn check_allowances(&self, owner: String) -> Result<JsValue, GuiError> {
+        let all_vaults = self.get_all_vaults();
```
hardyjosh, line 139: instead of checking allowances for all vaults, shouldn't we be checking allowances for all deposits?

we don't need to check allowance for input vault tokens (that's for tokens coming in, so you don't deposit into an input vault). we also don't need to check allowance for output vault tokens that aren't being deposited into.

crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    fn get_deposits_as_map(&self) -> HashMap<Address, String> {
+        self.get_deposits()
+            .into_iter()
+            .map(|d| (d.address, d.amount))
+            .collect()
+    }
+
+    async fn get_token_decimals(&self, rpc_url: &str, token: Address) -> Result<u8, GuiError> {
```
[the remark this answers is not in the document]
findolor: yeah this makes a lot of sense since we can just use that in other places in the application. will add that

crates/js_api/src/gui/order_operations.rs
findolor: ok i wasn't sure before implementing this but i get the picture now, will do it based on this

crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    async fn get_deposit_calldatas(
+        &self,
+        vaults_and_deposits: &Vec<(OrderIO, U256)>,
+    ) -> Result<Vec<Vec<u8>>, GuiError> {
+        let mut results = Vec::new();
+        for (order_io, amount) in vaults_and_deposits.iter() {
+            let deposit_args = DepositArgs {
+                token: order_io.token.address,
+                vault_id: order_io.vault_id.unwrap_or(rand::random()),
```
hardyjosh, line 176: why is this needed?

crates/js_api/src/gui/order_operations.rs
```diff
@@ -0,0 +1,279 @@
...
+            let allowance = self
+                .check_allowance(&orderbook, &deposit_args, &owner)
+                .await?;
+            results.push(allowance);
+        }
+
+        Ok(serde_wasm_bindgen::to_value(&results)?)
+    }
+
+    /// Generate approval calldatas for all inputs and outputs of the order
```
hardyjosh, line 129: still says inputs?

hardyjosh (approved):

Issue #1035, opened 2024-12-01
use strictyaml for parsing dotrain yaml


hardyjosh:
as part of #965, we want to switch over to using strictyaml for our dotrain order frontmatter.

https://docs.rs/crate/strict-yaml-rust/latest

the current implementation attempts to parse our yaml into more complex types, with some undesired effects that we can't really control. whereas strictyaml only has strings, arrays and maps.

this means we'll need to do all additional parsing ourselves.

this implementation probably looks something like:
- removing `ConfigSource` and `Config` entirely
- instead, creating a `OrderbookYaml` and `DotrainOrderYaml` struct ([see the spec](https://github.com/rainlanguage/specs/blob/main/ob-yaml.md))
- these will use strictyaml to create a flat struct that matches the structure of the spec (i.e. in a `Deployment` the `order` field is a string with the name of the order, not an `Order`). other types would be parsed (i.e. a `Vault` `vault-id` would be `U256`.
- these structs would be able to take a `Vec`of yamls that would be merged (i.e. a `DotrainOrderYaml` could be composed with an `OrderbookYaml`, like we do now with 'settings' and orders)
- implementing getters that will retrieve nested data on-demand [according to the spec](https://github.com/rainlanguage/specs/blob/main/ob-yaml.md)
- implementing setters that will update nested data (i.e. updating a gui deployment's order's first input's token's address)
- a serialisation function to convert the `DotrainOrderFrontmatter` struct back into a string, such that deserialising would produce the same struct
- a validation function (?)

the goals for this revised implementation are:
- explicit parsing of the yaml into types other than strings
- a controlled entrypoint into the data, so for example we can do on the fly token replacement using `${order.input[0].token.address}` syntax and know it will be enforced everywhere
- one source of truth when working with frontmatters (at the moment having both `ConfigSource` and `Config` causes issues where we have to retain both)
- always being able to go back to a string without losing any data integrity

other considerations:
- strictyaml doesn't contain anchors/refs, so we can't use that functionality as we were thinking. but it was pretty awkward anyway (ie required stringly merging yamls which was getting kinda hacky)
- some of this could be done incrementally, for example `plots` could be re-emitted as a string and parsed using the existing `serde_yaml` implementation

PR #1043, opened 2024-12-04
Creating new OrderbookYaml and DotrainYaml structs

findolor:
## Motivation

See issue: #1035

## Solution

- Created two new structs called `OrderbookYaml` and `DotrainYaml`
- Implemented the constructor function that converts a yaml string to each struct
- Added tests

## Checks

## Missing things
- [x] Plot support in charts field in `DotrainYaml`
- [ ] Data validation during initialization and setter functions
- [ ] Field validation during initialization and setter functions
- [ ] Getter functions for each struct


crates/settings/src/network.rs
```diff
@@ -45,9 +47,81 @@ impl Network {
...
+                    rpc: Url::from_str(&require_string(
+                        &value,
+                        Some("rpc"),
+                        Some(format!("rpc string missing in network: {key}")),
+                    )?)
+                    .map_err(|e| ParseNetworkConfigSourceError::RpcParseError(e))?,
+                    chain_id: require_string(
+                        &value,
+                        Some("chain-id"),
+                        Some(format!("chain-id string missing in network: {key}")),
```
hardyjosh, line 77: i think saying string here is a bit confusing, because it's also an error in the next step if it's a string (because it can't be parsed to a u64). from the user perspective if they put "one", that's not valid, but this error indicates it would be

crates/settings/src/network.rs
```diff
@@ -45,9 +47,81 @@ impl Network {
...
+            Some("missing field: networks".to_string()),
+        )?;
+
+        let networks = networks_hash
+            .into_iter()
+            .map(|(key, value)| {
+                let key = key.as_str().unwrap_or_default().to_string();
+                let network = Network {
+                    document: document.clone(),
+                    name: key.clone(),
```
hardyjosh, line 67: i know it's not part of this PR, but i think the name field should be renamed to "key" as it's more generic and can be used across all the structs, whereas "name" collides with Token name for example

crates/settings/src/network.rs
```diff
@@ -45,9 +47,81 @@ impl Network {
         }
     }
 
+    pub fn parse_networks_from_yaml(
```
hardyjosh, line 50: if we make this generic like `parse_all_from_yaml` could it be trait?

crates/settings/src/network.rs
```diff
@@ -45,9 +47,81 @@ impl Network {
...
+            Some("networks"),
+            Some("missing field: networks".to_string()),
+        )?;
+
+        let networks = networks_hash
+            .into_iter()
+            .map(|(key, value)| {
+                let key = key.as_str().unwrap_or_default().to_string();
+                let network = Network {
+                    document: document.clone(),
```
hardyjosh, line 66: if you clone the document it will still hold the reference back to the original right?

PR #1049, opened 2024-12-09
Clear2 0 order ratio bug

ninokeldishvili:
## Motivation
There seems to be a bug in an edge with clear2() where if Alice or Bob order ratio is 0, then the tx goes through silently without any reverts or issues, but nothing actually clears, no vault balance changes and no bounties.

This should not happen, in fact if the order ratio is 0, that measn all the output vault should go to the tx sender as bounty.

## Solution
Add test showing the issue 

## Checks


src/concrete/ob/OrderBook.sol
```diff
@@ -786,14 +786,18 @@ contract OrderBook is IOrderBookV4, IMetaV1_2, ReentrancyGuard, Multicall, Order
...
+            // solidity generic index out of bounds errors.
             if (calculateOrderStack.length < CALCULATE_ORDER_MIN_OUTPUTS) {
                 revert UnsupportedCalculateOutputs(calculateOrderStack.length);
             }
 
-            Output18Amount orderOutputMax18 = Output18Amount.wrap(calculateOrderStack[1]);
-            uint256 orderIORatio = calculateOrderStack[0];
+            uint256 orderIORatio;
+            Output18Amount orderOutputMax18;
+            assembly ("memory-safe") {
```
thedavidmeister, line 797: this was done to make the contract size smaller so that it can be deployed, we're right at the max codesize limit

rouzwelt (approved):

src/concrete/ob/OrderBook.sol
```diff
@@ -786,14 +786,18 @@ contract OrderBook is IOrderBookV4, IMetaV1_2, ReentrancyGuard, Multicall, Order
                 new uint256[](0)
             );
 
-            // This is redundant with the array index checks implied by solidity
-            // but it's a much clearer error message.
+            // This is a much error message and overall more efficient than
```
hardyjosh, line 789: typo?

PR #1082, opened 2024-12-16
1060 withdraw2 task not being executed when withdrawamount is 0

### LINKED ISSUE #1060: withdraw2 task not being executed when withdrawAmount is <= 0
atm, withdraw2 task is inside of an `if` block that checks if withdrawAmount is > 0, however bot relies on the task to ensure the bounty of the transaction, and because sometime the bounty can be 0 for one of the tokens, that ends up bypassing the task and as a result, bot fails to ensure the bounty check correctly resulting in negative bounty txs. 

The withdraw2 task used to be outside of the `if` block, but since has been moved inside of it.

ninokeldishvili:
## Motivation
atm, withdraw2 task is inside of an if block that checks if withdrawAmount is > 0, however bot relies on the task to ensure the bounty of the transaction, and because sometime the bounty can be 0 for one of the tokens, that ends up bypassing the task and as a result, bot fails to ensure the bounty check correctly resulting in negative bounty txs.

## Solution

Write test to show the issue

## Checks


thedavidmeister:
👤 human
Ruled e60119b19c38692dfa770ca68e1264891ecd804c: close-candidate — Closing as SUPERSEDED — nothing here needs salvaging, checked rather than assumed.

This PR was never a fix. Its own Solution section says "Write test to show the issue": +101/-1, one test file, a REPRODUCTION of the bug reported in #1060. The bug it demonstrates was fixed by the change that moved the withdraw event and doPost out of the `if (withdrawAmount > 0)` guard, and #1060 was closed on 2026-08-05 after verifying that fix on main directly (src/concrete/raindex/RaindexV6.sol:291-327 has no amount guard at all; doPost is gated solely on `post.length != 0`).

Every behaviour this PR reaches for is already covered on main, by tests that discriminate harder:

- Its PURPOSE — the task runs when the withdraw amount lands at 0 — is pinned by `test/concrete/raindex/RaindexV6.withdraw.entask.t.sol:291 testRaindexWithdrawalEvalZeroAmountEvalNoop`, which deposits `Float.wrap(0)` (empty vault), bounds the target positive, runs `:ensure(0 "always fails")` and REQUIRES the harness to observe that revert, at 100 fuzz runs. This PR's `assert(writes.length > 0)` is a weaker proxy for the same claim — state writes could originate elsewhere.
- Its CONTEXT assertions are a subset of `testOrderWithdrawContext` on main, which asserts EIGHT words against fuzzed inputs — `raindex()`, `withdrawer()`, `withdraw-token()`, `withdraw-vault-id()`, `withdraw-vault-before()`, `withdraw-vault-after()`, `withdraw-target-amount()`, `withdraw-token-decimals()`. Main splits this PR's single `withdraw-vault-balance()` into before AND after, so the balance delta is pinned rather than only the pre-state, and adds token decimals, which this PR does not cover.

It also could not pass as written. It deposits `depositAmount >= 1` and withdraws `targetAmount >= 1`, then asserts `withdraw-amount() == 0`; under the fixed code the actual amount is `min(target, balance)`, which is positive, so that ensure fails. The inconsistency is not a defect in the author's work — the test was written against the BUGGY code to demonstrate the bug, and `ci: nochecks` means it never ran. It is written against `OrderBook.withdraw2` and `test/concrete/ob/`, both since renamed to `RaindexV6.withdraw4` and `test/concrete/raindex/`, so it does not compile against main either.

Recorded because the pattern is worth catching earlier: this is the rain.dia#6 shape — an open PR and a close-candidate flag each holding half the picture, neither queue reading the other. #1060's flag correctly did not treat this PR as a blocker (it cited a landing, and a redundant PR in flight does not un-land what landed), but closing the issue left this PR orphaned with `closes: [1060]` pointing at a closed issue. Disposing of it is the PR-lane half of that same ruling.

PR #1153, opened 2025-01-14
Add Sidebar and Homepage to webapp

hardingjam:
## Checks


hardyjosh:
I think we need a different version of this for the webapp because we'll have different pages - ie no add order and no settings

hardyjosh:
@hardingjam the idea of the ui-components lib is to have shared components. SidebarWebapp should just be Sidebar, but in the webapp. Scoping a shared component back to a specific one doesn't make sense.

hardyjosh:
@hardingjam same for Homepage. We'll have a different homepage for the webapp so it doesn't make sense to move it into the shared lib. Would be better if this PR just dealt with Sidebar. Then we can handle the homepage separately.

hardyjosh (changes requested):
See comments

hardyjosh (approved):

Issue #1203, opened 2025-01-22
Serialize state errors if no tokens are in the yaml


hardingjam:
Now that we add tokens manually in some cases, we need to be able to serialize state when they arent yet in the yaml

PR #1212, opened 2025-01-23
Refactoring state serialization/deserialization

findolor:
## Motivation

See issue: #1203

## Solution

- Removes current deployment dependency from serialization
- Turn deserialization function into a static one
- Add new parsing methods for gui and order
- Check for dotrain hash and return error if not valid

## Checks


crates/settings/src/gui.rs
```diff
@@ -337,6 +337,78 @@ impl Gui {
...
+                            }
+                        } else {
+                            return Err(YamlError::ParseError(format!(
+                                "fields list missing in gui deployment: {deployment_key}"
+                            )));
+                        }
+                    }
+                } else {
+                    return Err(YamlError::ParseError(
+                        "deployments field must be a map in gui".to_string(),
```
hardyjosh, line 405: the idea that we're looking for a map with a key foo, and if we find foo but it's a list, seems like a reusable thing and the error could have the key foo replaced in it expected {foo} to be a map, found a list instead

hardyjosh (changes requested):
see comment

hardyjosh (approved):

PR #1245, opened 2025-02-07
Create CheckboxMyItemsOnly.svelte

ninokeldishvili:
## Motivation
Rather than the accounts dropdown, we should just have a single checkbox "Only show my orders/vaults"

We still want someone to be able to uncheck that filter and see all orders, but filtered by connected wallet would be the default (ie checked on).

If a wallet isn't connected, it will show all orders, but when the wallet is connected it should check the filter on.

If the user switches the account in their wallet, the filtered list should update.

## Solution
Implement filtering by connected account

## Checks


hardyjosh:
If a wallet isn't connected, I can still check the checkbox, but nothing happens

image

if no wallet is connected, can we disable the checkbox, with a tooltip on hover that says "Connect a wallet to filter by order owner"

hardyjosh (approved):

Issue #1309, opened 2025-02-16
Order Page does reflect the trade and the balance change instantly.


Siddharth2207:
For this order : 
https://v2.raindex.finance/orders/base-0x0f0382c9cd9b3b62e96bec3c9cdfc70f6bccc2192944fa5a962e2e0294474f93

I can see that the order is traded but the balance change isn't reflected in the vault section and in the trades chart, but can see a trade has gone through.
<img width="1259" alt="Image" src="https://github.com/user-attachments/assets/0ac834f1-b1bc-4a9b-8925-66ef3e225d34" />

Can see once the page is refreshed : 
<img width="1250" alt="Image" src="https://github.com/user-attachments/assets/cad535c9-9b5a-44f2-8f4e-d1df9e710057" />

hardyjosh:
related to #1303

PR #1329, opened 2025-02-17
1303 order page isnt polling and can be out of sync

### LINKED ISSUE #1303: order page isn't polling and can be out of sync
the order page isn't polling for updates, we could poll every 15s

we could also have a refresh for the whole order page (refresh icon top right)

plus, everything should be synced. if someone hits refresh for the order/quote/trades, it should invalidate the query key for the whole order

otherwise you can get a situation like this, where the trade has gone through, but the vault balances don't reflect that

Image

hardingjam:
## Motivation

Closes #1303 and #1309 

## Checks


hardyjosh:
@hardingjam this still doesn't quite work - if i hit the refresh on the trades table, i can get a new trade without the vault balances being updated.

all of the refresh buttons on the page should be fully synced, there isn't really a use case for something on the page refreshing while something else doesn't - it's all related data

hardyjosh:
@hardingjam still doesn't work, I tried hitting the trade refetch and expected to see the quotes refetch but that isn't happening

hardyjosh (approved):

PR #1344, opened 2025-02-18
Update field descriptions on select token change

### LINKED ISSUE #1338: Field descriptions not updating on token change
If for some reason you input the token addresses and then change them (lets say because you want to swap buy/sell) the titles of the form do not update

Image

findolor:
## Motivation

See issue: #1338

## Checks

fix  #1338


packages/ui-components/src/lib/components/deployment/DeploymentSteps.svelte
```diff
@@ -186,6 +188,15 @@
 
 	async function _handleUpdateGuiState(gui: DotrainOrderGui) {
 		await areAllTokensSelected();
+
+		if (allTokensSelected) {
+			let newAllTokenInfos = await gui.getAllTokenInfos();
+			if (newAllTokenInfos !== allTokenInfos) {
```
hardyjosh, line 194: hm this isn't a deep comparison of equality, does it actually work? this is something we should show in a test

also, we are calling getAllTokenInfos with every state change, but we only need to do that if the tokens have changed

so we need more granular control over our state change callback

findolor: I think this will be easier to implement after merging #1361. Let's visit this again after that PR

hardyjosh (approved):

PR #1387, opened 2025-02-27
Change order links to orderHash not order IDs

hardingjam:
- This raises another issue about whether all order id use in rust should be hash instead. Since all sub-components of OrderDetail  currently require the ID.

## Checks


hardyjosh:
@hardingjam we don't use orderhash as the ID in the subgraph because you could conceivably have two orders on two chains with the same orderhash

hardyjosh (approved):

Issue #1623, opened 2025-04-09
Missing test - Common crate - `add_order.rs`


findolor:
- [ ] `AddOrderArgs::execute` is not tested
- [ ] `AddOrderArgs::get_add_order_calldata` is not tested

PR #1722, opened 2025-05-02
Add tests for add order file

findolor:
## Motivation

See issue: https://github.com/rainlanguage/rain.orderbook/issues/1623

## Checks

- ~[ ] included screenshots (if this involves a front-end change)~

fix https://github.com/rainlanguage/rain.orderbook/issues/1623

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive tests covering rainlang parsing, composition, error handling, and order calldata generation.
  - Improved validation of order-related functionalities including vault ID assignment and RPC URL handling.
  - Removed an obsolete test and introduced a helper function to streamline test setups.


crates/common/src/add_order.rs
```diff
@@ -974,4 +978,225 @@ _ _: 16 52;
...
+        };
+        let rainlang = add_order_args.compose_to_rainlang().unwrap();
+        assert_eq!(
+            rainlang,
+            "/* 0. calculate-io */ \n_ _: 10 20;\n\n/* 1. handle-io */ \n:;"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_add_order_call_try_into_write_contract_parameters() {
```
0xgleb, line 1151: I'm a bit confused - is this testing any logic from this module?

crates/common/src/add_order.rs
```diff
@@ -974,4 +978,225 @@ _ _: 16 52;
...
+                0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 21, 2, 0, 0, 0, 12, 2, 2, 0, 2, 1, 16, 0, 0, 1,
+                16, 0, 0, 0, 0, 0, 0
+            ]
+        );
+    }
+
+    #[tokio::test]
+    async fn test_compose_to_rainlang() {
```
0xgleb, line 1102: we should probably also cover unhappy path of `compose_to_rainlang`

crates/common/src/add_order.rs
```diff
@@ -974,4 +978,225 @@ _ _: 16 52;
             .await
             .expect_err("expected to fail but resolved");
     }
+
+    #[tokio::test]
+    async fn test_try_parse_rainlang() {
```
0xgleb, line 983: we should probably also cover unhappy path of `try_parse_rainlang`

crates/common/src/add_order.rs
0xgleb, line 1: I took a look at the tests that were in the module prior to the PR and noticed a couple of things

- Only happy path of `try_generate_meta` is covered
- Only happy path of `compose_addorder_post_task` is covered
- `get_add_order_calldata` is not covered
- `execute` is not covered, although it relies on `LedgerClient` so only an error case can be covered
- `test_order_config_v2_validity` doesn't seem to do anything

0xgleb (changes requested):

crates/common/src/add_order.rs
findolor: i have added this test because we cannot mock the ledger logic. this was the function that was called inside the `execute` function so i have added this to test out to flow inside that function

0xgleb: maybe it'd be better to take that bit out of `execute`, put it in a new function and test that? otherwise the logic in `execute` can diverge from the test and the test wouldn't catch any potential issues

hardyjosh (approved):

PR #1735, opened 2025-05-06
Do not reset transaction store when modals are closed

hardingjam:
## Motivation

Closing the transaction modals resets the state of transactionStore, this means that anything "listening" for updates to transactions (eg Toasts) cannot track their state once modals are closed.

## Solution

Instead of resetting the transactionStore when the transactionModal is closed, reset it when a new transaction is initiated.

## Checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Transaction state now resets only when starting new transactions, not when closing modals or cancelling actions.
  - Closing or cancelling transaction-related modals no longer resets the transaction state unexpectedly.

- **Tests**
  - Updated tests to confirm that cancelling a deposit or withdrawal does not trigger a transaction state reset.
  - Removed tests that asserted transaction state resets on modal closure.


packages/webapp/src/__tests__/OrderRemoveModal.test.ts
```diff
@@ -48,23 +48,4 @@ describe('OrderRemoveModal', () => {
...
-		render(OrderRemoveModal, defaultProps);
-		const resetSpy = vi.spyOn(transactionStore, 'reset');
-
-		const closeButton = screen.getByLabelText('Close modal');
-		await fireEvent.click(closeButton);
-
-		expect(resetSpy).toHaveBeenCalled();
-	});
-
-	it('calls onRemove callback after successful transaction', async () => {
```
findolor, line 62: Why did you remove this test? `onRemove` is still used inside the component and we don't have test for it now

findolor (changes requested):

packages/webapp/src/__tests__/OrderRemoveModal.test.ts
hardingjam: It shouldn't actually be in the component at all. #1738 removes the need for onRemove
Having onRemove callbacks in the modal doesn't make sense since once the modal is closed, it will no longer respond to the transactionEvents.

I've added a TransactionsListener to handle invalidate queries and showing Toasts when transactions succeed, but that raises an additional issue with how transactions work, #1743

hardyjosh (approved):

findolor (approved):

PR #2084, opened 2025-08-14
feat: Add price ratio functionality to DotrainOrderGuide

### LINKED ISSUE #1998: Ability to to reference price ratios in a field names/descriptions and for default settings in DotrainOrderGui yaml
Our auction DCA and DSF strats require an `initial-io` which helps the strat start auctions at or near the current market price.

This is a pain for users and often something they get wrong, sometimes leading to total loss of funds. They have to find the right ratio and make sure it's in the right direction.

In our custom GUI's (e.g. Cyclo), we simply use the sushiswap lib to pull the current market price and the end user doesn't need to do anything.

Ideally, strategy writers would be able to define where external prices should be used as part of the YAML config, and the SDK would handle fetching the ratio and inserting it in the correct place. But, we don't want to enshrine a specific method of fetching prices.

The proposed solution would consist of:

1. A way of specifying where a price ratio should be used. We already have `${...}` to use contextual info, so it could simply be `${io-ratio(input, output)}`, e.g. `${io-ratio(order.inputs.0.address, order.outputs.0.address)}`. 

This could be used in a couple of places.

As a default:

```yaml
       - binding: initial-io
          name: Kickoff ${order.inputs.0.token.symbol} per ${order.outputs.0.token.symbol}
          description: |
            The initial ${order.inputs.0.token.symbol} per ${order.outputs.0.token.symbol} to kickoff the first auction.

            This ratio is calculated in the same way as the baseline ratio.

            It must be greater than the baseline ratio, regardless of what you are selling or buying.
            This is because getting more input per output is always better for you.
            The initial auction will start high (at the value you set here) and then drop to the baseline ratio over time.
            Subsequent auctions will start at some % above the last auction price and drop to the baseline ratio over time.
         default: ${io-ratio(order.inputs.0.address, order.outputs.0.address)}
```

or, in the name/description of a field (basically as a helper for the user inputting into the field.

```yaml
   base:
      name: Base
      description: Deploy a limit order on Base.
      deposits:
        - token: token2
      fields:
        - binding: fixed-io
          name: ${order.inputs.0.token.symbol} per ${order.outputs.0.token.symbol}
          description: Fixed exchange rate (${order.inputs.0.token.symbol} received per 1 ${order.outputs.0.token.symbol} sold). 
          The current market price is ${io-ratio(order.inputs.0.address, order.outputs.0.address)}
      select-tokens:
        - key: token1
          name: Token to Buy
          description: Select the token you want to purchase
        - key: token2
          name: Token to Sell
          description: Select the token you want to sell
```

2. The actual method of fetching prices injected as a dependency when creating the `DotrainOrderGui`, basically just a js function to be used as a callback.

```javascript
const priceCallback = (inputTokenAddress, outputTokenAddress) => {
  return fetchPrice(inputTokenAddress, outputTokenAddress)
};
const gui = DotrainOrderGui.newWithDeployment(dotrain, 'deployment', priceCallback);
```

We'd need to think through error handling etc.

big14way:
- Add support for ${io-ratio(input, output)} expressions in YAML configurations
- Extend Context system to parse and resolve price ratio expressions
- Add price callback support to DotrainOrderGui for external price fetching
- Implement comprehensive error handling for price callback failures
- Add backward-compatible API with newWithDeploymentAndPriceCallback method
- Include complete documentation and examples for price ratio usage
- Support dynamic field names, descriptions, and default values with market prices

Resolves #1998: Ability to reference price ratios in field names/descriptions and for default field values

## Summary by CodeRabbit

- New Features
  - Dynamic price ratio resolution in the GUI using an external price callback.
  - YAML support for ${io-ratio(...)} expressions across names, descriptions, and defaults.
  - Optional pricing support when creating the GUI and when restoring from saved state.
  - Interpolated default values for GUI fields.

- Documentation
  - Added README section, comprehensive guide, JavaScript examples, and a YAML example demonstrating dynamic pricing.

- Tests
  - Added tests for price resolution, callback handling, and error cases.

- Chores
  - Updated external dependency reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,40 @@ We use wasm-bindgen to create the `orderbook` package from our Rust crates, whic
 
 This same package is [published to npm](https://www.npmjs.com/package/@rainlanguage/orderbook), allowing developers to more easily create their own frontends for Raindex.
 
+## New Features
+
+### 🎯 Dynamic Price Ratio Support
+
+The DotrainOrderGui now supports dynamic price ratio resolution using `${io-ratio(input, output)}` expressions in YAML configurations. This allows strategy writers to reference real-time market prices without hardcoding values.
+
+**Key Benefits:**
+- **Automatic Price Fetching**: Strategies can automatically use current market prices
+- **User-Friendly Defaults**: No more manual price lookups for users
+- **Flexible Integration**: Works with any price API (SushiSwap, 1inch, CoinGecko, etc.)
+- **Error Handling**: Comprehensive error handling for network issues and invalid data
+
+**Quick Example:**
+```yaml
+fields:
+  - binding: initial-io
+    name: Kickoff ${order.inputs.0.token.symbol} per ${order.outputs.0.token.symbol}
+    description: Current market price: ${io-ratio(order.inputs.0.address, order.outputs.0.address)}
+    default: ${io-ratio(order.inputs.0.address, order.outputs.0.address)}
+```
+
+```javascript
+const priceCallback = async (inputTokenAddress, outputTokenAddress) => {
+  const response = await fetch(`/api/price/${inputTokenAddress}/${outputTokenAddress}`);
+  return (await response.json()).ratio.toString();
+};
+
+const gui = await DotrainOrderGui.newWithDeploymentAndPriceCallback(
+  yaml, deployment, stateCallback, priceCallback
+);
+```
+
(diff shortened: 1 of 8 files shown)


findolor (changes requested):
@big14way Thanks for the PR. We use AI analysis to apply improvements and catch issues so could you take a look at all the AI comments posted in this PR and reply to them?

You don't have to agree with every comment but sometimes it outlines issues that have been overlooked. If you can sweep and reply to each (not needed/will do) that would be great. I can review the PR after.

PR #2280, opened 2025-10-29
Update bootstrap pipeline to run for engine and runner stages

findolor:
## Motivation

- The current bootstrap implementation was only working for the engine stage which ran for all the orderbooks. With the reset logic it had, it was dropping all the tables and data for all the networks even though a single network/orderbook needed a reset.

## Solution

- Split run into two stages - `engine_run` and `runner_run`
- Implement runner for the client, making sure we do db resets properly
- Implement engine logic to only reset a specific orderbook data instead of the whole db
- Update and tests

## Checks

## Summary by CodeRabbit

* **New Features**
  * Separated bootstrap functionality into specialized engine and runner entry points for improved operation handling.
  * Added new error types with enhanced error messages for better diagnostics.

* **Bug Fixes**
  * Improved data validation with stronger type handling for numeric fields and addresses.

* **Refactor**
  * Streamlined database query execution and ordering logic.
  * Converted internal data representations to more reliable type systems.
  * Simplified error handling pathways for clearer failure reporting.

### DIFF
--- a/crates/cli/src/commands/local_db/sync/config.rs
+++ b/crates/cli/src/commands/local_db/sync/config.rs
@@ -49,8 +49,6 @@ orderbooks:
     network: arbitrum
     subgraph: arbitrum
     deployment-block: "123"
-local-db-remotes:
-  arbitrum: https://example.com/localdb/arbitrum
 "#;
 
     #[test]
--- a/crates/cli/src/commands/order/add.rs
+++ b/crates/cli/src/commands/order/add.rs
@@ -140,9 +140,6 @@ deployers:
         network: some-network
         address: 0xF14E09601A47552De6aBd3A0B165607FaFd2B5Ba
 
-local-db-remotes:
-    some-orderbook: https://example.com/localdb/arbitrum
-
 orderbooks:
     some-orderbook:
         address: 0xc95A5f8eFe14d7a20BD2E5BAFEC4E71f8Ce0B9A6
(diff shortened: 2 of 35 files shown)


findolor:
@coderabbitai review

crates/common/src/local_db/decode.rs
```diff
@@ -215,9 +216,56 @@ fn decode_event<E: SolEvent>(event: &LogEntryResponse) -> Result<E, DecodeError>
...
+
+    let original = events.to_vec();
+    for (position, (idx, _, _)) in keyed.into_iter().enumerate() {
+        events[position] = original[idx].clone();
+    }
+
+    Ok(())
+}
+
+fn parse_u64_hex_or_dec(value: &str) -> Result<u64, std::num::ParseIntError> {
```
hardyjosh: I feel like I've seen this before - as a rule I think it would be good to put generic helpers like this in common/utils

crates/common/src/local_db/pipeline/adapters/bootstrap.rs
```diff
@@ -113,16 +114,30 @@ pub trait BootstrapPipeline {
...
     ) -> Result<(), LocalDbError>
     where
         DB: LocalDbQueryExecutor + ?Sized,
     {
-        Err(LocalDbError::MissingBootstrapImplementation)
+        db.query_text(&clear_orderbook_data_stmt(ob_id)).await?;
+        Ok(())
+    }
+
+    async fn engine_run<DB>(&self, _: &DB, _: &BootstrapConfig) -> Result<(), LocalDbError>
```
hardyjosh: do you actually need to do this or can you just not do a default implementation for these fns?

findolor: the engine implementation for bootstrap is different for both client and producer so we can't use default. are you saying this should be a function in env specific structs and not in the trait itself or something else?

hardyjosh (approved):

PR #2421, opened 2026-01-28
Raindex Router Arb contract

rouzwelt:
## Motivation
This PR adds Raindex Router Arb contract, which can perform routed `takeOrders()` by doing a flash loan and making a intermediate swap through external source (or even can be another raindex order although this PR only implements the sushi RP handler), here is a simple example:

- Assume we have orders A/B and wanna trade it, but there is no exact mirror order (B/A) and external liquidity sources (such as Sushi or Balancer) also don't have such pool and offer
- We have another order with C/A
- There is some route in external source for B -> C
- We can trade these 2 order against each other, first taking a flash loan of token A to take the first order, we will get B as a result, next swap B through external sources for market price and get C, then take the last order and complete the circuit
- Pay back the flash loan
- Take whatever profit left from A and C tokens (there cant be any token B profits since it all is swapped to C through external source)

Note: The maximum clearable amount in this case cannot exceed `total orderbook balance of token A / 2`, thats because we cant take all the orderbook token A balance as loan since we need them for the last `takeOrder()` to complete, example:
Assume the order C/A has deposited 100 A tokens, and thats all the A tokens orderbook has, so at most half of it can be flash loaned, which goes towards taking the first order and next half that stayed in orderbook for taking the last order

## Checks

## Summary by CodeRabbit

* **New Features**
  * Flash-loan driven arbitrage router for cross-orderbook arbitrage flows
  * On-chain multi-leg swap routing with Sushi leg support for dynamic asset conversion
  * ARB-network deployment support for the new router

* **Bug Fixes**
  * Guarded order processing to avoid division-by-zero in IO ratio calculations

* **Tests**
  * End-to-end tests covering arbitrage execution, vault accounting, and balance assertions


thedavidmeister:
**Flag for Protofire (audit scoping):** the high-severity OrderBookV6 zero-IO-ratio div-by-zero this PR addresses is ALREADY on main with a mutation-validated test (RaindexV6.sol zero-amount guard at the take/market-sell path; test/concrete/raindex/RaindexV6.takeOrder.freshMutation.t.sol::testZeroRatioSkips) — main SKIPS a zero-ratio order (OrderZeroAmount) before the division, so no fix is needed there. What remains in THIS PR is ~430 lines of NEW, unaudited router/arb contracts (RaindexRouterOrderBookV6Arb.sol, OrderBookV6RaindexRouter.sol + Deploy.sol). Those are NOT in the bytecode being audited now and must be SCOPED SEPARATELY by Protofire if/when landed — do not assume the orderbook audit covers them.

PR #2435, opened 2026-02-04
vaultless trade bug

rouzwelt:
This PR adds a test for vautless trade (`takeOrder4()`),
this test currently fails most probably due to the fact that:

- `recordVaultIO()` for the order tries to first `pushTokens()` to the order owner for the order's input by calling `increaseVaultBalance()`.
- This potentially can succeed fi orderbook has enough balance to cover for that token transfer, but if not, this direct token transfer fails due to the fact that the required tokens have not been taken from the taker yet 
- Why this doesnt fail for taking normal orders? because there is no direct token transfer in `increaseVaultBalance()` and `increaseVaultBalance()`, its just a change in the vault balance mappings in that case

## Checks


thedavidmeister:
This reproduces audit finding **M01 (#2618)** — vault-0 input push-before-pull. The code here is pre-rename (`OrderBookV6`) / pre-soldeer and no longer builds against `main` (CI is fully red). I've ported it to a current-`RaindexV6` repro in **#2644** (adds the ambient-balance success case and a `clear3` case). Closing in favour of #2644; the underlying bug remains open as #2618.

PR #2547, opened 2026-04-14
feat: SignedContextInjector trait for caller-supplied signed contexts

hardyjosh:
## Motivation

There is currently no way for a caller of the take-orders flow to contribute its own `SignedContextV1` entries per order. The only source of signed contexts today is the `RaindexSignedContextOracleV1` URL discovered from order meta — fetched by the library via HTTP POST. That model assumes a single, publicly-reachable oracle URL per order, which doesn't fit callers that need to produce per-request, per-user signatures in-process (for example, an API that reserves specific orders for authenticated clients by issuing a per-request gating signature).

## Solution

- New `SignedContextInjector` trait in `crates/quote/src/injector.rs`:
  ```rust
  async fn contexts_for(&self, order: &OrderV4, input_io_index: u32, output_io_index: u32, counterparty: Address) -> Vec<SignedContextV1>;
  ```
  Default `NoopInjector` preserves existing behaviour. Placed in the quote crate because common depends on quote, not the reverse; re-exported from `rain_orderbook_common::take_orders` for API parity.
- Threaded through `get_order_quotes_batch_with_injector` so composed contexts land on `QuoteV2.signedContext` **before** the quote RPC. Strategies whose `calculate-io` asserts on signed context now quote successfully instead of silently being filtered as failed quotes.
- `RaindexOrderQuote` gains `signed_context: Vec<SignedContextV1>` (serde-default, tsify-optional, skip-if-empty) so candidates inherit the exact context the quote saw — no post-quote re-attach.
- `build_take_order_candidates_for_pair` forwards `counterparty: Address` + `&dyn SignedContextInjector`.
- `get_take_orders_calldata` stays backward-compatible (defaults to `NoopInjector` + `Address::ZERO`). Added `get_take_orders_calldata_with_injector` as the injector-aware entry point; deliberately not `#[wasm_export]`ed since trait objects don't serialize.
- Composition order is `[oracle..., injected...]`. Strategies that read contexts positionally (`signer<N>()`, `signed-context<N M>()`) must follow this convention.
- Minor behavioural change: oracle fetches now receive the real `counterparty` rather than the hardcoded `Address::ZERO` stub. Existing oracle servers ignore this field (per the removed comment), so this is a no-op in practice but removes a misleading value.

## Checks

Verified locally:
- `cargo check --workspace` clean
- `cargo test -p rain_orderbook_common --lib take_orders` — 176 passed
- `cargo test -p rain_orderbook_quote --lib` — 48 passed
- `cargo fmt`

## Summary by CodeRabbit

* **New Features**
  * Quotes now include signed context data and accept counterparty information.
  * Added an extensible injector interface (and a no-op default) to append custom signed contexts during quote generation.
  * New APIs allow callers to provide an injector when requesting quotes or take-calldata.

* **Refactor**
  * Candidate/quote pipeline now composes contexts as [oracle..., injected...] up-front and preserves them through selection and execution (removes deferred oracle fetch).

* **Tests**
  * Serialization and quote tests updated to validate signed context behavior.

### DIFF
--- a/crates/common/src/raindex_client/mod.rs
+++ b/crates/common/src/raindex_client/mod.rs
@@ -462,6 +462,8 @@ pub enum RaindexError {
     QuoteDataMissing,
     #[error("Invalid input index: {0}")]
     InvalidInputIndex(u32),
+    #[error("Invalid output index: {0}")]
+    InvalidOutputIndex(u32),
     #[error("Cannot parse metadata: {0}")]
     ParseMetaError(#[from] rain_metadata::Error),
     #[error("No metaboards configured for any chain")]
@@ -654,6 +656,12 @@ impl RaindexError {
             RaindexError::InvalidInputIndex(index) => {
                 format!(
                     "Invalid input index: {}. The order does not have an input at this index.",
+                    index
+                )
+            }
+            RaindexError::InvalidOutputIndex(index) => {
+                format!(
+                    "Invalid output index: {}. The order does not have an output at this index.",
                     index
                 )
             }
--- a/crates/common/src/raindex_client/order_quotes.rs
+++ b/crates/common/src/raindex_client/order_quotes.rs
@@ -1,9 +1,13 @@
 use super::*;
 use crate::raindex_client::orders::RaindexOrder;
 use crate::raindex_client::orders_list::RaindexOrders;
+use alloy::primitives::Address;
 use rain_math_float::Float;
-use rain_orderbook_bindings::IRaindexV6::OrderV4;
-use rain_orderbook_quote::{get_order_quotes, BatchOrderQuotesResponse, OrderQuoteValue, Pair};
+use rain_orderbook_bindings::IRaindexV6::{OrderV4, SignedContextV1};
+use rain_orderbook_quote::{
+    get_order_quotes, BatchOrderQuotesResponse, NoopInjector, OrderQuoteValue, Pair,
+    SignedContextInjector,
+};
 use rain_orderbook_subgraph_client::utils::float::{F0, F1};
 use std::ops::{Div, Mul};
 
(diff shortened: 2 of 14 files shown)


hardyjosh:
@coderabbitai review

hardyjosh:
@coderabbitai full review

hardyjosh:
@coderabbitai review

hardyjosh:
@coderabbitai full review

findolor (changes requested):
Single-order take looks out of sync with the new signed-context flow. getTakeCalldata still quotes via get_quotes(), which hardcodes Address::ZERO + NoopInjector, and build_candidate_from_quote() drops any quote.signed_context before execution. That means batch take is now counterparty/injector-aware, but single take is not. For gated orders that depend on signed context during calculate-io, single take can still fail at quote time or lose injected context before calldata generation.

hardyjosh:
@findolor good catch — addressed in d48f48e3.

**Fixes:**
1. `build_candidate_from_quote` now inherits `quote.signed_context` instead of dropping it. Oracle and injected contexts composed at the batch-quote layer now propagate all the way through single-take.
2. Added `RaindexOrder::get_quotes_with_injector` and `RaindexOrder::get_take_calldata_with_injector` (non-wasm) that thread `counterparty` + `&dyn SignedContextInjector` through the single-order path. Wasm-exported `get_quotes` / `get_take_calldata` stay on `NoopInjector + Address::ZERO` for backward compat.
3. Removed the now-stale oracle re-fetch inside `execute_single_take` (it was oracle-only and would have clobbered any injected entries the candidate already carried). Dropped `oracle_url` from `TakeOrderExecutionParams` since it was the only consumer.

New unit test `test_build_candidate_inherits_signed_context_from_quote` asserts the candidate inherits both oracle and injected entries with correct ordering.

hardyjosh:
CodeRabbit round 2 addressed in 5a1d92ae:

- **Defer `signed_context` clone until candidate survives filtering** — moved the clone into `try_build_candidate`'s final `Ok(Some(...))` branch, dropping the up-front clone on every quote. Updated tests.
- **Explicit `InvalidOutputIndex` error on `get_take_calldata`/`get_take_calldata_with_injector`** — added the variant to `RaindexError` with a readable message and guard in both methods so invalid caller indices are no longer indistinguishable from `NoLiquidity`.

findolor (approved):

hardyjosh:
### Merge activity

* **Apr 15, 3:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/rainlanguage/raindex/2547).
* **Apr 15, 3:44 PM UTC**: @hardyjosh merged this pull request with [Graphite](https://app.graphite.com/github/pr/rainlanguage/raindex/2547).

PR #2611, opened 2026-06-01
test: finalizeArb zero-balance path + fuzz (audit GAP-A12-1/A12-4)

thedavidmeister:
Audit **#2537** (`LibRaindexArb.finalizeArb`), test-only.

- **GAP-A12-1** — zero-leftover path: when the arb leaves zero token balance (raindex pulls exactly the swapped amount), `finalizeArb`'s `if (balance > 0)` guards skip the `safeTransfer`s. Covered by a `pull == swap` case (call succeeds, sender gets nothing, arb empty).
- **GAP-A12-4** — fuzz: vary swap/pull amounts and assert `finalizeArb` sweeps exactly the input-token profit (`swap - pull`) to the sender and leaves the arb empty, across the full range including the zero-leftover boundary.

No source change, so no pointer regen. Both pass (fuzz 5096 runs).

Closes #2537


thedavidmeister:

PR #2612, opened 2026-06-01
test: RouteProcessor arb receive()/fallback() accept ETH (audit A07-2)

thedavidmeister:
Audit **#2534** A07-2 (`RouteProcessorRaindexV6ArbOrderTaker`), test-only.

The arb has `receive() external payable {}` and `fallback() external payable {}` (so a route processor can refund native gas, which `finalizeArb` later sweeps), but the bare ETH-receipt paths had no coverage.

- `testReceiveAcceptsEth` — empty calldata → `receive()` accepts ETH; arb holds it. (fuzzed amount)
- `testFallbackAcceptsEth` — one-byte (non-selector) calldata → `fallback()` accepts ETH; arb holds it. (fuzzed amount)

A07-3 (fuzz of the full `onTakeOrders2`/`arb5` cycle) is a follow-up — it needs careful Float/wei handling and is left for a separate change.

Addresses #2534 (A07-2).


thedavidmeister:

PR #2613, opened 2026-06-01
test: LibRaindex.doPost edge paths (audit GAP-A11-1/2/3)

thedavidmeister:
Audit **#2536** (`LibRaindex.doPost`), test-only — covers the three guarded paths via a harness + `vm.mockCall`/`mockCallRevert`:

- **GAP-A11-1** — empty `post` array is a no-op (loop never runs).
- **GAP-A11-2** — a task with empty `bytecode` is skipped; `eval4` and `store.set` are both mocked to revert if reached, and `doPost` completes cleanly.
- **GAP-A11-3** — `store.set` is called only when `eval4` produces writes: the no-write case (set mocked to revert) completes without calling it; the with-write contrast asserts `store.set` *is* called.

No source change. All 4 pass.

Closes #2536


thedavidmeister:

PR #2616, opened 2026-06-02
feat: add ERC4626 read support

findolor:
## Summary
- add minimal IERC4626 and IMulticall3 bindings for ERC4626 read support
- add raindex_common::erc4626::ERC4626 single-vault helpers plus batch share-ratio reads
- batch ERC4626 vault reads through Multicall3 pinned to a captured block, returning block number, optional block timestamp, captured_at, and per-vault results/errors
- wire ERC4626 errors into RaindexError for Rust REST-side usage

## Multicall design
This uses direct Multicall3 aggregate3 encoding instead of Alloy's multicall builder because each batch phase mixes heterogeneous return types across a runtime-sized vault list. Alloy supports heterogeneous multicall well for compile-time tuples, and dynamic runtime-sized multicall for one call type at a time. Using Alloy here would require splitting the read into more Multicall3 eth_calls by return type, while the direct aggregate3 path keeps the batch to fewer pinned calls and still preserves per-token failures.

## Testing
- cargo fmt --all
- cargo test -p raindex_common erc4626 --lib
- cargo check -p raindex_common
- nix develop .#wasm-shell -c rainix-rs-static

## Notes
- Rust downstream crates can import raindex_common::erc4626::ERC4626 directly.
- Single-token callers can use ERC4626::batch_share_ratios with a one-item vault list.
- Base Multicall3 is used by default, with an optional override address.
- JS/WASM SDK export is intentionally out of scope for this PR.

## Summary by CodeRabbit

* **New Features**
  * Added ERC-4626 vault protocol support including asset address queries, decimal information, and share-to-asset conversions
  * Introduced batch vault querying capabilities with optimized RPC-based multicall operations for improved efficiency
  * Extended error handling framework with comprehensive vault-specific failure diagnostics

### DIFF
--- a/crates/bindings/src/lib.rs
+++ b/crates/bindings/src/lib.rs
@@ -41,6 +41,34 @@ sol!(
 );
 
 sol!(
+    #![sol(all_derives = true, rpc)]
+    interface IERC4626 {
+        function asset() external view returns (address);
+        function decimals() external view returns (uint8);
+        function convertToAssets(uint256 shares) external view returns (uint256);
+        function convertToShares(uint256 assets) external view returns (uint256);
+    }
+);
+
+sol!(
+    #![sol(all_derives = true)]
+    interface IMulticall3 {
+        struct Call3 {
+            address target;
+            bool allowFailure;
+            bytes callData;
+        }
+
+        struct Result {
+            bool success;
+            bytes returnData;
+        }
+
+        function aggregate3(Call3[] calldata calls) external payable returns (Result[] memory returnData);
+    }
+);
+
+sol!(
     #![sol(all_derives = true)]
     #![sol(extra_derives(serde::Serialize, serde::Deserialize))]
     IInterpreterStoreV3, "./abis/IInterpreterStoreV3.json"
--- a/crates/common/src/erc4626.rs
+++ b/crates/common/src/erc4626.rs
@@ -1,0 +1,1329 @@
+use crate::{erc20, utils::amount_formatter};
+use alloy::eips::{BlockId, RpcBlockHash};
+use alloy::network::{AnyNetwork, TransactionBuilder};
+use alloy::primitives::{address, Address, B256, U256};
+use alloy::providers::Provider;
+use alloy::rpc::types::TransactionRequest;
+use alloy::serde::WithOtherFields;
+use alloy::sol_types::{Revert, SolCall, SolError};
+use alloy::transports::TransportError;
+use rain_error_decoding::{AbiDecodeFailedErrors, AbiDecodedErrorType};
+use raindex_bindings::provider::{mk_read_provider, ReadProvider, ReadProviderError};
+use raindex_bindings::IERC4626::IERC4626Instance;
+use raindex_bindings::{
+    IERC20Metadata::decimalsCall as erc20DecimalsCall,
+    IMulticall3::{aggregate3Call, Call3, Result as Multicall3Result},
+    IERC4626::{assetCall, convertToAssetsCall, decimalsCall as erc4626DecimalsCall},
+};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use url::Url;
+#[cfg(target_family = "wasm")]
+use wasm_bindgen_utils::{impl_wasm_traits, prelude::*};
+
+#[cfg(not(target_family = "wasm"))]
+use std::time::{SystemTime, UNIX_EPOCH};
+
+pub const DEFAULT_MULTICALL3_ADDRESS: Address =
+    address!("cA11bde05977b3631167028862bE2a173976CA11");
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(target_family = "wasm", derive(Tsify))]
+#[serde(rename_all = "camelCase")]
+pub struct ERC4626ShareAssetConversion {
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub share_token_address: Address,
+    pub share_token_decimals: u8,
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub asset_address: Address,
+    pub asset_decimals: u8,
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub shares: U256,
+    pub shares_display: String,
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub assets: U256,
+    pub assets_display: String,
+}
+#[cfg(target_family = "wasm")]
+impl_wasm_traits!(ERC4626ShareAssetConversion);
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(target_family = "wasm", derive(Tsify))]
+#[serde(rename_all = "camelCase")]
+pub struct ERC4626BatchVault {
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub vault_address: Address,
+    #[cfg_attr(target_family = "wasm", tsify(optional, type = "string"))]
+    pub shares: Option<U256>,
+    #[cfg_attr(target_family = "wasm", tsify(optional, type = "string"))]
+    pub expected_asset_address: Option<Address>,
+}
+#[cfg(target_family = "wasm")]
+impl_wasm_traits!(ERC4626BatchVault);
+
+impl ERC4626BatchVault {
+    pub fn new(vault_address: Address) -> Self {
+        Self {
+            vault_address,
+            shares: None,
+            expected_asset_address: None,
+        }
+    }
+
+    pub fn with_shares(mut self, shares: U256) -> Self {
+        self.shares = Some(shares);
+        self
+    }
+
+    pub fn with_expected_asset_address(mut self, expected_asset_address: Address) -> Self {
+        self.expected_asset_address = Some(expected_asset_address);
+        self
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(target_family = "wasm", derive(Tsify))]
+#[serde(rename_all = "camelCase")]
+pub struct ERC4626BatchItem {
+    #[cfg_attr(target_family = "wasm", tsify(type = "string"))]
+    pub vault_address: Address,
+    pub success: bool,
+    #[cfg_attr(target_family = "wasm", tsify(optional))]
+    pub data: Option<ERC4626ShareAssetConversion>,
+    #[cfg_attr(target_family = "wasm", tsify(optional))]
+    pub expected_asset_matches: Option<bool>,
+    #[cfg_attr(target_family = "wasm", tsify(optional))]
+    pub error: Option<String>,
+}
+#[cfg(target_family = "wasm")]
+impl_wasm_traits!(ERC4626BatchItem);
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(target_family = "wasm", derive(Tsify))]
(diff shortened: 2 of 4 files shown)


findolor:
* **#2616** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2616).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/bindings/src/lib.rs
```diff
@@ -40,6 +40,34 @@ sol!(
...
+    #![sol(all_derives = true, rpc)]
+    interface IERC4626 {
+        function asset() external view returns (address);
+        function decimals() external view returns (uint8);
+        function convertToAssets(uint256 shares) external view returns (uint256);
+        function convertToShares(uint256 assets) external view returns (uint256);
+    }
+);
+
+sol!(
```
0xgleb, line 53: doesn't alloy have built-in multicall bindings?

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+        let error = match (error, data.is_none()) {
+            (Some(error), _) => Some(error),
+            (None, true) => Some("Incomplete ERC4626 batch result".to_string()),
+            (None, false) => None,
+        };
+
+        ERC4626BatchItem {
+            vault_address: self.vault_address,
+            success: error.is_none(),
```
JuaniRios, line 438: should fix: `success` is just `error.is_none()`, so a vault whose on-chain `asset()` differs from `expected_asset_address` still comes back `success: true` with full data, and only `expected_asset_matches: Some(false)` signals the problem. A caller that checks `success`/`data` (the natural path) treats a wrong-underlying vault as a clean read, which is the wrong-data case this field is meant to catch. Either gate `success` on the asset match, or document that callers must inspect `expected_asset_matches` separately, and add a test for the mismatch.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        .client()
+        .request::<_, Option<BlockResponse>>("eth_getBlockByNumber", params)
+        .await
+        .ok()
+        .flatten()
+        .and_then(|block| block.timestamp.try_into().ok())
+}
+
+fn captured_at_unix_timestamp() -> u64 {
+    SystemTime::now()
```
JuaniRios, line 652: should fix: `SystemTime::now()` panics on `wasm32-unknown-unknown`, and this crate compiles for wasm (the new types derive Tsify / impl_wasm_traits). The repo already avoids this elsewhere with `js_sys::Date::now()` under `target_family = "wasm"`, see `utils/timing.rs` and `quote/order_quotes.rs`. It is not reachable from JS yet because the SDK export is out of scope, but it is a landmine for that export and a convention break. Split this by target and use `js_sys::Date::now() / 1000.0` (guarded) on wasm.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+    pub async fn batch_share_ratios(
+        rpcs: Vec<Url>,
+        vaults: Vec<ERC4626BatchVault>,
+        multicall3_address: Option<Address>,
+    ) -> Result<ERC4626BatchResponse, Error> {
+        let captured_at = captured_at_unix_timestamp();
+        let provider = mk_read_provider(&rpcs)?;
+        let block_number = provider.get_block_number().await?;
+        let block_id = BlockId::Number(BlockNumberOrTag::Number(block_number));
```
JuaniRios: should fix: the snapshot is pinned by block number, but `mk_read_provider` builds a FallbackLayer with `active_transport_count = rpcs.len()`, so with multiple RPCs each request can hit a different node. A block number is not a stable identity across nodes at the head or across a reorg, so the reported block/timestamp and the per-vault reads can come from different states while presenting as one snapshot. Consider fetching the block once and pinning every `eth_call` to `BlockId::Hash`, or documenting that single-RPC is required for consistency.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        let block_timestamp = get_block_timestamp(&provider, block_number).await;
+        let multicall3_address = multicall3_address.unwrap_or(DEFAULT_MULTICALL3_ADDRESS);
+
+        let mut states: Vec<BatchState> = vaults.iter().map(BatchState::new).collect();
+
+        if !states.is_empty() {
+            read_phase1(&provider, block_id, multicall3_address, &mut states).await?;
```
JuaniRios, line 226: should fix: a transport failure here, or a Multicall3 address with no code on the target chain, fails the whole batch via `?` and discards all per-vault state. The description says the design preserves per-token failures, but that only holds for per-sub-call reverts through `allowFailure`, not batch-level failures. Worth documenting that `Err` means batch-infra failure (distinct from per-vault `items[n].error`), and ideally distinguishing a missing or undecodable Multicall3 from a per-vault revert, since the override address implies non-canonical chains are expected.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        assert_eq!(data2.share_token_address, vault2);
+        assert_eq!(data2.asset_address, asset2);
+        assert_eq!(data2.asset_decimals, 6);
+        assert_eq!(data2.shares, shares);
+        assert_eq!(data2.assets, U256::from(1_500_000u64));
+        assert_eq!(data2.assets_display, "1.5");
+    }
+
+    #[tokio::test]
+    async fn test_batch_share_ratios_per_item_failure() {
```
JuaniRios, line 917: should fix: the failure-path coverage is one-sided. This only tests a phase-1 `asset()` failure, so phase-2's error/skip machinery is untested, and there is no test for `expected_asset_matches == Some(false)` (which ties to the success-gating comment above, a regression making it always `Some(true)` would pass here), an empty vault list, or the `MalformedMulticallResponse` length guard. Adding those would cover the riskiest logic in this module.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        let shares = vault.convertToShares(assets).call().await;
+
+        match shares {
+            Ok(shares) => Ok(shares),
+            Err(err) => Err(handle_alloy_err(err, "Convert to shares reverted").await),
+        }
+    }
+
+    pub async fn share_ratio(&self) -> Result<ERC4626ShareAssetConversion, Error> {
+        let share_decimals = self.share_decimals().await?;
```
JuaniRios, line 300: minor: `share_ratio` reads `share_decimals()` here, then `convert_shares_to_assets` reads it again, so every call does two identical decimals RPCs. Passing the already-fetched value through a private helper would drop the second round-trip. Separately, `build_conversion` exists twice (this method and the free fn lower down) byte-for-byte except the address source, so the method could just delegate to the free fn.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        let one_share = U256::from(10).pow(U256::from(share_decimals));
```
JuaniRios: minor: `10 ** share_decimals` uses ruint `pow`, which wraps on overflow in release. `decimals()` is a uint8 so 78 through 255 are ABI-valid, and `10^78` exceeds U256::MAX, so an odd vault produces a wrapped share amount that gets sent to `convertToAssets` and then fails formatting later. No real vault does this, but `checked_pow` plus an explicit unsupported-decimals error would be cleaner. Same pattern at the batch default around line 439.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        source,
+    })
+}
+
+fn decode_multicall_return<C: SolCall>(
+    result: &Multicall3Result,
+    label: &str,
+) -> Result<C::Return, String> {
+    if !result.success {
+        return Err(format!("{label} call failed: {}", result.returnData));
```
JuaniRios: minor: this emits raw revert bytes as hex, while the single-vault path (`handle_alloy_err`) decodes revert data into readable messages. For the common `require(cond, "msg")` pattern the batch error is unreadable. A best-effort `Error(string)` decode (selector 0x08c379a0) with a hex fallback would match the single-vault error quality.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+    let block_hex = format!("0x{block_number:x}");
+    let params = serde_json::json!([block_hex, false]);
+
+    provider
+        .client()
+        .request::<_, Option<BlockResponse>>("eth_getBlockByNumber", params)
+        .await
+        .ok()
+        .flatten()
+        .and_then(|block| block.timestamp.try_into().ok())
```
JuaniRios: minor: any transport error, missing block, or failed U256 to u64 conversion all collapse to `None` here, so a caller cannot tell a recoverable RPC error from a chain that genuinely has no timestamp. Folding the timestamp into a single block fetch (which also helps the hash-pinning point above) and distinguishing error from absent would be more robust.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+    aggregate3Call::abi_decode_returns(&bytes).map_err(|source| Error::SolTypesError {
+        msg: "Failed to decode Multicall3 aggregate3 return".to_string(),
+        source,
+    })
+}
+
+fn decode_multicall_return<C: SolCall>(
+    result: &Multicall3Result,
+    label: &str,
+) -> Result<C::Return, String> {
```
JuaniRios: minor: this returns `Result<_, String>` built with `format!`, and the string is stored and surfaced as the item error, even though the file defines a structured thiserror `Error` enum just below. The per-item-must-be-serializable constraint is fair, but the decode layer is where the type info is lost, and a test now asserts on the exact wording ("asset call failed"), making the message a load-bearing contract. A small serializable `ItemError` enum, stringified only at the boundary, would avoid that.

crates/common/src/erc4626.rs
```diff
@@ -0,0 +1,1041 @@
...
+        let server = MockServer::start_async().await;
+
+        server.mock(|when, then| {
+            when.method("POST")
+                .path("/rpc")
+                .body_contains(encode_prefixed(erc4626DecimalsCall::SELECTOR));
+            then.json_body(json!({
+                "jsonrpc": "2.0",
+                "id": 1,
+                "result": "0x1",
```
JuaniRios, line 1322: nit: this returns `0x1` and asserts `is_err()`, but a `decimals()` of 1 is a valid vault. It passes only because `0x1` is not a valid 32-byte ABI-encoded uint8, so it is really testing alloy's decoder rejecting a malformed response, not this module. Rename it to reflect that, or point it at a vault that reverts with a real error selector and assert the decoded message.

JuaniRios (changes requested):

crates/bindings/src/lib.rs
findolor: Yeah, we looked at Alloy’s multicall support. It works well for compile-time heterogeneous tuples, and for dynamic batches when the calls are the same return type. Here each runtime-sized batch phase mixes return types, so using Alloy’s builder would split this into more Multicall3 `eth_call`s by return type. The direct `aggregate3` encoding keeps fewer pinned calls while still preserving per-vault failures. We’re still using Alloy `sol!` call types for ABI encoding/decoding.

JuaniRios (approved):

findolor:
Closing this because the ERC4626 implementation belongs in rain-erc instead. Replacement PR: https://github.com/rainlanguage/rain.erc/pull/32

PR #2627, opened 2026-06-02
fix: audit H01 — thread calculate-phase state across same-owner batch orders

### LINKED ISSUE #2617: Audit (Protofire H01): takeOrders4 defers calculate-phase kvs — repeated orders in a batch bypass stateful limits
**Severity:** High

### H01 - takeOrders4 delays calculate-phase state persistence, allowing repeated-order batches to bypass stateful order limits

In takeOrders4(), each order is evaluated in the main processing loop through calculateOrderIO(). That evaluation returns kvs from the calculate entrypoint, but those key-value writes are not persisted immediately. Instead, they are stored on the in-memory OrderIOCalculationV4 struct and only written later in handleIO(), after the main order-processing loop has already finished. The calculation evaluation is performed with an empty stateOverlay, so repeated evaluations in the same transaction do not see the calculation-phase writes from prior iterations in that batch. Because takeOrders4() does not enforce uniqueness in the config.orders array, a taker can submit the same order multiple times, e.g., [orderA, orderA, orderA]. Each repeated calculateOrderIO() call will observe the same persisted interpreter state, even though earlier iterations in the batch may already have logically “used up” part of a strategy’s intended allowance. Vault balances are still updated immediately, so this does not bypass vault-balance caps, but interpreter-state-based calculate limits are deferred and therefore not visible to later calculations in the same batch.

**Path:** src/concrete/ob/OrderBookV6.sol

**Impact:** Orders that rely on interpreter state in the calculate entry point to enforce cumulative limits, such as per-block, per-epoch, or rolling execution caps, can be over-consumed in a single takeOrders4 transaction by simply repeating the same order in the batch. This can allow a taker to bypass stateful strategy limits and drain the order up to the remaining vault balance.

**Recommendation:** Persist calculate-phase kvs before evaluating the next order in the batch, or pass the prior calculate writes forward through stateOverlay so later evaluations see them. If deferred persistence must remain, then enforce uniqueness of order hashes within a single takeOrders4() call.

**Found in:** e686b4d3

---
- `src/concrete/ob/OrderBookV6.sol` → `src/concrete/raindex/RaindexV6.sol`.
- Confirmed present on `main`: `calculateOrderIO` accumulates into `orderIOCalculationsToHandle`; `handleIO` runs after the take loop; no per-batch order-hash uniqueness.

thedavidmeister:
## Audit Protofire H01 (#2617) — High

`takeOrders4` evaluated each order's calculate entrypoint with an empty `stateOverlay` and deferred its `kvs` to `handleIO` (after the whole order loop). With no per-batch uniqueness, repeating the same order in one call made every calculate evaluation re-read the same stale store state, so a stateful per-order limit enforced in the calculate phase (per-block / per-epoch / rolling caps) is bypassed within a batch.

## Fix
Thread each evaluated order's calculate-phase `kvs` into later orders' calculate evals via `stateOverlay`, **scoped per owner namespace**: the interpreter caches store reads by raw key (the namespace only applies on the persisted-store fallback), so a flat batch-wide overlay would leak state between different owners. `clear3` / `quote` (single-order) pass an empty overlay.

- `buildBatchStateOverlay` concatenates prior same-owner kvs (later writes win).
- `BatchCalculateState` + helper extraction (`calculateOrderIOBatched`, `capOutputMaxToVaultBalance`) keep the loop within stack limits.

## Tests
- `RaindexV6.takeOrder.batchStateBypass.t.sol`
  - `testH01RepeatedOrderBypassesCalculatePhaseCap` — a "fill once" order repeated 3× in one batch caps at 1 (was 3 before the fix).
  - `testH01CrossOwnerStateIsolation` — two different owners using the same key each fill independently (proves the per-owner scoping doesn't leak).

## Gas (measured A/B via `gasleft`; `--gas-report` can't profile the etched orderbook)
| Scenario | Before | After | Δ |
|---|---|---|---|
| Single order (common) | 146,943 | 148,776 | +1,833 (+1.2%) |
| 5 same-owner stateful | 574,975 | 603,668 | +28,693 (+5.0%) |

Overhead is bookkeeping only (no new SLOAD/SSTORE in the common path); it scales with same-owner repeats (the abusive shape the fix neutralises).

## Contract size
RaindexV6 runtime **24,493 / 24,576** (83 B margin) — still deploys. (The `forge --sizes` "exceeds" error is `RaindexV6VaultBalanceHarness`, a test-only `vm.etch`'d subclass, not the deployed contract.)

## ⚠️ Deploy follow-up
The fix changes RaindexV6 bytecode, so the deterministic address moves (`0x83507d7E…` → `0x3Ee54EbB…`). Only `RaindexV6.pointers.sol` is regenerated here; merging will need the same redeploy + metadata cascade as #2610 (arb pointers, `LibRaindexDeploy` start blocks, `networks.json`, `subgraph.yaml`, ABIs) before CI's prod-deploy / drift checks pass.

Closes #2617.

## Release Notes

* **New Features**
  * Enhanced batch order processing with improved state threading for same-owner orders
  * Extended order calculation to support state overlay input

* **Tests**
  * Added comprehensive test coverage for batch order execution behavior
  * New tests verify state isolation across different order owners within a single batch


thedavidmeister:
### Gas impact (measured)

`--gas-report` can't profile the orderbook here (the test harness `vm.etch`'s RaindexV6 from pointers rather than deploying it), so this is a clean A/B: `gasleft()` around `takeOrders4` for scenarios that behave **identically** before/after the fix, so the delta is pure overhead. Pre-fix = `origin/main`; post-fix = this branch.

| Scenario | Before | After | Δ | % |
|---|---|---|---|---|
| Single order (common case) | 146,943 | 148,776 | +1,833 | +1.2% |
| 5 same-owner orders writing store state (overlay worst case) | 574,975 | 603,668 | +28,693 | +5.0% |

**Breakdown**
- The added work is memory bookkeeping only — the `BatchCalculateState` record, `buildBatchStateOverlay`, and two helper frames. No new `SLOAD`/`SSTORE` on the common path.
- Single order / all-distinct-owner batches: overlay is empty, so the interpreter does zero extra seeding → low fixed overhead (~1.8k gas).
- Same-owner repeats: each order copies prior same-owner kvs into its overlay and the interpreter re-seeds them → ~O(N²·K), the +5% above for N=5. This is the abusive shape the fix exists to neutralise (those orders now hit `outputMax==0` and skip when a real cap is in play).

Contract size after the fix: RaindexV6 runtime **24,493 / 24,576 B** (83 B margin) — still deployable.

thedavidmeister:
### Adversarial test coverage + mutation validation

Added `RaindexV6.takeOrder.batchStateAdversarial.t.sol` (8 tests) on top of the 2 in `…batchStateBypass.t.sol`. Each test was **mutation-checked** — the relevant fix logic was deliberately broken (pointers regenerated each time) and the claimed-covering tests were confirmed to fail, with everything else staying green:

| Mutation | Tests that fail (and only these) | Line covered |
|---|---|---|
| **M1** `buildBatchStateOverlay` → return empty | RepeatedCap, ThreadingAcrossDistinct, Skipped, RatioSkipped, Cumulative | overlay threading |
| **M2** drop the `owners[j] == owner` filter | CrossOwnerStateIsolation, CrossOwnerNoLeakInterleaved | per-owner scoping |
| **M3** record only non-zero-output orders | Skipped (only) | recording for `OrderZeroAmount` skips |
| **M4** record only orders that would fill | Skipped + RatioSkipped | recording precedes *both* skip checks |

No test passes under a mutation it claims to catch, and the isolations are clean (M3 uniquely caught by the zero-amount test; M4 adds the ratio-skip path).

**Eventualities covered:** repeated-order cap; threading across *distinct* same-owner orders; both skip paths (`OrderZeroAmount`, `OrderExceedsMaxRatio`) still thread their calculate writes; cross-owner isolation incl. interleaved owners; cumulative counters (values, not just a flag); cross-transaction persistence unchanged; key-specific reads; a dead order (`OrderNotFound`) interleaved without corrupting the accumulator.

**Covered by design (not separately tested):** `clear3`/`quote` pass an empty overlay (single-eval, distinct namespaces); `maximumIO` exhausted mid-batch simply stops evaluating later orders (no calculate runs → nothing to thread).

thedavidmeister:
Reviewed 7a58895a4 — H01 fix (per-owner stateOverlay threading) reviewed; 10 tests mutation-validated (M1–M4 each fail exactly the claiming tests); RaindexV6 + flash-borrower redeployed to the new deterministic addresses; deploy metadata (start blocks / networks.json / subgraph.yaml) set to the exact on-chain creation blocks. CI 17/17 green.

PR #2637, opened 2026-06-03
ci: temporary soldeer push diagnostic

thedavidmeister:
Throwaway diagnostic to capture the real `forge soldeer push` error in CI (RUST_LOG=trace + endpoint reachability). Removed after we read the result.


thedavidmeister:
Reviewed 00b430997: throwaway workflow_dispatch diagnostic only (reachability curl + RUST_LOG=trace soldeer push); no production code, safe to merge.

PR #2641, opened 2026-06-03
ci: curl soldeer upload workaround

thedavidmeister:
Tests a curl-based multipart upload to the soldeer registry, bypassing soldeer 0.10.0's broken push. If it publishes 0.1.0, this becomes the autopublish fix.


thedavidmeister:
Reviewed 589003de8: throwaway workflow_dispatch test of a curl multipart upload (soldeer 0.10.0 push workaround). No production code, token via secret only. Safe.

PR #2642, opened 2026-06-03
ci: remove throwaway soldeer diagnostic

thedavidmeister:
0.1.0 is published, autopublish is green; removing the temporary diagnostic workflow.


thedavidmeister:
Reviewed efdea4e2a: removes the throwaway soldeer diagnostic workflow only. Safe.

PR #2645, opened 2026-06-04
ci: install soldeer deps before subgraph-deploy

thedavidmeister:
The `Deploy subgraph` workflow (`deploy-subgraph.yaml`) failed on `main`: `subgraph-deploy` runs a `forge build`, but the workflow never installs the soldeer dependencies, so every import (`forge-std`, `@openzeppelin`, `rain-*`, `raindex-interface`, …) resolves to a missing file — >256 errors, aborting, exit 1.

Pre-existing breakage from the soldeer migration that never updated this workflow (last successful subgraph deploy was Mar 24, pre-migration). Fix: run `forge soldeer install` + `forge build` before `subgraph-deploy`, matching the README/AGENTS bootstrap.

Validated by dispatching the deploy from this branch (carries main's contracts).

### DIFF
--- a/.github/workflows/deploy-subgraph.yaml
+++ b/.github/workflows/deploy-subgraph.yaml
@@ -31,6 +31,10 @@ jobs:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
           restore-prefixes-first-match: nix-${{ runner.os }}-
           gc-max-store-size-linux: 1G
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed a99d05eb2: approved — adds `forge soldeer install` + `forge build` before `subgraph-deploy` so the deploy's forge build resolves its dependencies (the missing install was the root cause of the >256-error abort). One-line CI-only change, all checks green, and validated by the successful deploy run 26944414936.

PR #2646, opened 2026-06-04
feat: pin + enforce 0.1.0 deployment constants for all raindex contracts

thedavidmeister:
Adds `_<version>`-suffixed deployed-address and runtime-codehash literals to `LibRaindexDeploy` for all six deployed contracts, recording the published `0.1.0` tag's deployment:

| contract | address (0.1.0) | codehash |
|---|---|---|
| RaindexV6 | `0xb05D73E6…3e3cE` | `0xf5d6441f…` |
| RaindexV6SubParser | `0xBdb9cC5F…dFF53` | `0x59e436b1…` |
| RouteProcessor4 | `0x6E2d0e71…68d21` | `0xeb3745a7…` |
| GenericPoolRaindexV6ArbOrderTaker | `0xA7299e32…4D0eb` | `0xba5ce714…` |
| RouteProcessorRaindexV6ArbOrderTaker | `0x3dbef2Fb…5216d` | `0x728605da…` |
| GenericPoolRaindexV6FlashBorrower | `0x8bdEd5aC…fA36c` | `0xa1bf69e2…` |

The canonical `*_DEPLOYED_ADDRESS` constants are derived from `src/generated/*.pointers.sol`, so they advance to new deterministic addresses every time the bytecode changes. These pinned literals keep a permanent reference to the `0.1.0` deployment.

## Enforcement

Adds `LibRaindexDeployTaggedConstants.t.sol` + `script/check-published-deploy-constants.sh`: an FFI test that queries the **live soldeer registry** for raindex's published versions and asserts `LibRaindexDeploy` pins a full constant suite (address + codehash for all six contracts) for each one. Publishing a new tag without pinning its constants now fails this test. It **skips** (rather than fails) if the registry is unreachable, so it never breaks on network flakiness.

Verified: passes with `0.1.0` fully pinned, and reports the exact `MISSING` constant when one is removed. All 12 pointer values pinned exactly; no pointer/codehash drift; `LibRaindexDeploy.t.sol` stays 31/31.

### DIFF
--- a/script/check-published-deploy-constants.sh
+++ b/script/check-published-deploy-constants.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+# SPDX-License-Identifier: LicenseRef-DCL-1.0
+# SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+#
 # Prints "OK" iff every version published to the soldeer registry for `raindex`
 # has a full suite of pinned deploy constants in LibRaindexDeploy.sol: a
 # DEPLOYED_ADDRESS and DEPLOYED_CODEHASH (suffixed with the version) for each of
--- a/script/check-published-deploy-constants.sh
+++ b/script/check-published-deploy-constants.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Prints "OK" iff every version published to the soldeer registry for `raindex`
+# has a full suite of pinned deploy constants in LibRaindexDeploy.sol: a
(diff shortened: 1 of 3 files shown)


thedavidmeister:
Reviewed 0783c27d6: approved — pins 0.1.0 deploy constants (address + codehash) for all six contracts (12/12 verified to match the committed pointers, no bytecode/pointer drift), plus an FFI test that enforces every published soldeer tag has its full constant suite (verified: passes today, reports the exact MISSING constant under mutation, skips if the registry is unreachable). All checks green.

PR #2655, opened 2026-06-05
ci: factor the webapp Vercel build into script/vercel-build.sh

thedavidmeister:
## What

Extract the duplicated inline webapp build steps from `vercel-preview.yaml` and `vercel-prod.yaml` into a single `script/vercel-build.sh`, and have both workflows call it.

## Why — a convention rainix can reference

`script/vercel-build.sh` is a **fixed-path convention**: it produces the Vercel *prebuilt* output for `packages/webapp`, run from the repo root with `PUBLIC_WALLETCONNECT_PROJECT_ID` exported. A rainix Vercel reusable workflow can then reference this path uniformly, so the **repo owns the build** and **rainix owns the generic wrapper** — nix install, the nix-store / cargo / npm caches, and the Vercel pull + deploy.

The build commands are extracted **verbatim** (same two `nix develop` steps, same cwd), so behaviour is unchanged.

## Scope note

This PR is the convention + de-duplication only. It does **not** add the cargo/npm caching that makes these deploys slow today — that intentionally lives in the rainix wrapper that will consume this script (the webapp Vercel workflows are not rainix-reusable-workflow jobs, so they inherit none of rainix's existing caching).

## Summary by CodeRabbit

* **Chores**
  * Consolidated build and deployment into a single centralized build entrypoint for more consistent, maintainable deployments.
  * Introduced caching of compiled artifacts and package dependencies to speed up builds and reduce CI time.


thedavidmeister:
Reviewed 0ddb50238: Refactor — extracts the inline webapp build from vercel-preview/prod into a single `script/vercel-build.sh` (the fixed-path contract the new `rainix-vercel` reusable references). Behavior preserved exactly: the script runs the same `.#wasm-shell` workspace build (`npm install --no-check` + the three `npm run build -w` libs) then the `packages/webapp` `.#webapp-shell` build, just relocated — the comment correctly notes the flake ref resolves as the prior inline step did. The workflows additionally gain `Swatinem/rust-cache` + npm cache (matching the reusable's caching) and keep `PUBLIC_WALLETCONNECT_PROJECT_ID` scoped to the build step. 17/17 checks green — preview/prod build CI actually ran the new script. Prerequisite for adopting rainix#220's `rainix-vercel` reusable (a follow-up can swap these jobs to `uses:` it). No blockers.

thedavidmeister:
Reviewed dc93f3e6b: Same diff as the previously-reviewed 0ddb50238 (head changed only because the branch was updated with main). Refactor extracting the inline webapp build into `script/vercel-build.sh` (the fixed-path contract for rainix#220's `rainix-vercel` reusable); build behavior preserved exactly (same `.#wasm-shell` workspace build + `packages/webapp` `.#webapp-shell` build), adds rust/npm caching, keeps `PUBLIC_WALLETCONNECT_PROJECT_ID` scoped to the build step. 17/17 checks green incl. the Deploy-Preview jobs that actually run the new script. No blockers. Admin-merging on explicit owner authorization (raindex requires an approving review the author can't self-provide).

PR #2644, opened 2026-06-03
fix: M01 vault-0 input push-before-pull (#2618)

thedavidmeister:
Fixes audit finding **M01 (#2618)** — _vaultId==0 on order input side requires pre-existing ambient balance (push-before-pull)_. Supersedes the stale #2435.

## Bug

`increaseVaultBalance` settles a `vaultId == 0` order INPUT as a direct `transfer` to the owner. The input push happened *before* the funds backing it were pulled in:
- **`takeOrders4`** pulls the taker's payment only after the per-order loop, but the vault-0 input push happened inside the loop.
- **`clear3`** ran Alice's input push before Bob's output (which funds it) was pulled.

So the orderbook transferred the input token to the owner while holding none of it — a logically fundable trade reverted with `ERC20InsufficientBalance` unless the orderbook happened to hold unrelated ambient balance.

## Fix — pull every incoming token before pushing any outgoing token

`recordVaultIO` split into three uniform phases, used by both `takeOrders4` and `clear3`:
- `recordVaultOutput` — decrease/pull (money in).
- `recordVaultInput` — credit internal input vaults only (vault 0 has no internal balance).
- `pushVaultZeroInput` — push vault-0 inputs out, run by each caller once the orderbook holds the token (`takeOrders4` after the taker pull; `clear3` after both outputs are pulled).

## Tests

`test/concrete/raindex/RaindexV6.takeOrder.vaultZeroInput.t.sol` uses real `MockToken` ERC20s with **zero ambient orderbook balance**, so they revert against the pre-fix bytecode and pass only once the orderbook is solvent at every push:
- `testVaultZeroInputSucceedsWithoutAmbientBalance` — `takeOrders4`
- `testClearVaultZeroInputSucceedsWithoutAmbientBalance` — `clear3`

Full Solidity suite: **424 pass**. Remaining reds are deploy-bookkeeping only (RPC-gated fork tests with no `*_RPC_URL`, and `testNetworksJsonAddresses` — see below).

## Deploy impact

Changing the bytecode changes the deterministic Zoltu/CREATE2 address:

| | before | after |
|---|---|---|
| Raindex | `0xb05D73E6…3e3cE` | `0x4A50fE00…a4CEC` |
| codehash | `0xf5d6441f…` | `0x9bd5ca20…` |

[...]

PR #2657, opened 2026-06-06
test: pin handle-IO sees the same order's calculate set (H01)

thedavidmeister:
## What

Add `testHandleIOSeesSameOrderCalculateSet` — pins the H01 invariant that an order's handle-IO sees its own calculate `:set` within the same evaluation.

## Why — a coverage gap found by mutation

`handleIO` commits the calculate-phase `kvs` to the store **before** running the handle-IO eval (contract comment: *"handle IO to see a consistent view on sets from calculate IO"*). That ordering was unpinned:
- *Dropping* the commit entirely → caught by ~18 tests.
- But **relocating** it to *after* the handle-IO eval (later orders still see the set; only the same order's handle-IO reads an unset `0`) → **survived the whole concrete suite**.

The new test fills an order whose calculate does `:set(0 5)` and whose handle-IO does `:ensure(equal-to(get(0) 5))` — it fills only if the ordering holds. Mutation-verified: under the relocate mutation it reverts (`hio sees calc`).

Part of a retroactive adversarial-mutation campaign over the H01 fix (#2617 / #2644). The rest of the H01 behaviours — calculate/handle-IO kvs commits, the maxRatio/zeroAmount untaken no-op skips, cross-owner isolation, same-owner threading, the repeated-order cap — are already densely covered.

## Summary by CodeRabbit

* **Tests**
  * Added test to verify correct execution ordering: order calculation phase data is committed to storage before the same order's input/output operations execute.

### DIFF
--- a/test/concrete/raindex/RaindexV6.takeOrder.batchStateAdversarial.t.sol
+++ b/test/concrete/raindex/RaindexV6.takeOrder.batchStateAdversarial.t.sol
@@ -324,4 +324,23 @@ contract RaindexV6TakeOrderBatchStateAdversarialTest is RaindexV6ExternalRealTes
         }
         assertEq(count, 1, "exactly one ContextV2 emitted per filled order");
     }
+
+    /// handleIO commits an order's calculate-phase `kvs` to the store BEFORE
+    /// running that order's handle-IO eval, so handle-IO sees its own calculate
+    /// `:set` within the same evaluation (the contract intent: "handle IO to see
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 1d19eac2c: H01 coverage-gap fill — pins that an order's handle-IO sees its own calculate `:set` (the step1-before-eval commit ordering), found by mutation (the relocate-after-eval mutant survived the whole suite). Discriminating + mutation-verified; test-only, no contract change; CI green 17/17. LGTM, merging.

PR #2659, opened 2026-06-06
docs: correct stale H01 docs in batchStateBypass.t.sol

### LINKED ISSUE #2658: Stale H01 docs in batchStateBypass.t.sol (describe removed stateOverlay + a non-existent bypass)
`test/concrete/raindex/RaindexV6.takeOrder.batchStateBypass.t.sol`'s documentation no longer matches the merged H01 implementation (#2644). The tests pass and are correct; the **comments** are stale and actively misleading.

## What's stale
- The file `@notice` says `takeOrders4` *"only persists the calculate-phase `kvs` later in `handleIO`, **after the whole order loop**"*, so *"a stateful per-order cap enforced in the calculate phase is **bypassed within a batch**"* (repeating an order 3× fills 3×).
- `testH01RepeatedOrderBypassesCalculatePhaseCap` carries the comment: *"This assertion … currently **FAILS (yields 3)** — it should pass once calculate-phase writes are **threaded forward via `stateOverlay`**."*

## What's actually true on current main (08d547fab)
- `handleIO` runs **per-iteration** inside the order loop, committing each order's calculate kvs **before the next order calculates** — not "after the whole order loop".
- The `stateOverlay` was **removed** by #2644 (`LibTakeOrdersOverlay` deleted); same-owner threading is via the store, not an overlay.
- The cap **holds**: `testH01RepeatedOrderBypassesCalculatePhaseCap` **passes** (total 1) — there is no within-batch bypass.

So the assertion is right, but the `@notice` + the *"currently FAILS … via stateOverlay"* comment describe a pre-fix/interim state and a mechanism that no longer exists. Suggest updating them (and reconsidering the `batchStateBypass` / `…Bypasses…` names) to the per-iteration `handleIO` mechanism.

Found during a retroactive adversarial-mutation campaign over the H01 fix; the campaign also added #2657 (a real coverage gap: handle-IO seeing the same order's calculate set).

thedavidmeister:
Closes #2658.

The `@notice` and test comments in `RaindexV6.takeOrder.batchStateBypass.t.sol` described persisting calculate `kvs` *after the whole order loop* via a `stateOverlay`, plus a within-batch cap *bypass* — none of which match the contract. `handleIO` runs **per order** and commits each order's calculate `kvs` to the store **before the next order calculates**, so the cap holds.

- Rewrote the `@notice` + the "fill once" / assertion / cross-owner comments to the current behaviour (removed the removed-`stateOverlay` references and the before/after narrative).
- Renamed `testH01RepeatedOrderBypassesCalculatePhaseCap` → `testH01RepeatedOrderCalculatePhaseCapHolds` (the cap holds; it doesn't bypass).

Test bodies unchanged; both pass. Surfaced by a retroactive adversarial-mutation campaign over the H01 fix (which also landed #2657).

## Tests
  * Updated test assertions for order batch processing to correctly validate per-order caps and state isolation behavior across batch operations.

### DIFF
--- a/test/concrete/raindex/RaindexV6.takeOrder.batchStateBypass.t.sol
+++ b/test/concrete/raindex/RaindexV6.takeOrder.batchStateBypass.t.sol
@@ -15,25 +15,26 @@ import {
 import {Float, LibDecimalFloat} from "rain-math-float-0.1.1/src/lib/LibDecimalFloat.sol";
 
 /// @title RaindexV6TakeOrderBatchStateBypassTest
-/// @notice Audit Protofire H01 (#2617): `takeOrders4` evaluates every order's
-/// calculate entrypoint with an empty `stateOverlay` and only persists the
-/// calculate-phase `kvs` later in `handleIO`, after the whole order loop. With
-/// no per-batch uniqueness, a taker can repeat the same order in one call and
-/// each calculate evaluation sees the same stale store state — so a stateful
-/// per-order cap enforced in the calculate phase is bypassed within a batch.
+/// @notice Audit Protofire H01 (#2617): a stateful per-order cap enforced in an
+/// order's calculate phase holds WITHIN a single `takeOrders4` batch, not only
+/// across transactions. Every order's calculate entrypoint evaluates with an
+/// empty `stateOverlay`, and `handleIO` runs per order inside the loop,
+/// committing that order's calculate-phase `kvs` to the store before the next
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 7aa93e717: docs-only — rewrites batchStateBypass.t.sol comments to current behaviour (drops the removed-stateOverlay references + before/after narrative) and renames the test to reflect the cap holds. Test bodies unchanged, both pass; CI green 16/16. Closes #2658. LGTM.

PR #2660, opened 2026-06-06
fix: arb4 requires IOIsInput == true (audit M03)

### LINKED ISSUE #2620: Audit (Protofire M03): arb4 computes flash-loan size from minimumIO without respecting IOIsInput
**Severity:** Medium

### M03 - arb4 computes flash-loan size from minimumIO without respecting IOIsInput semantics

In the flash-borrower path, arb4() determines the flash-loan amount by converting takeOrders.minimumIO using the borrowed token’s decimals. That only makes sense when minimumIO represents the taker input side of the trade. But in takeOrders4(), the meaning of minimumIO is controlled by IOIsInput: if IOIsInput is false, then minimumIO refers to taker output instead. In that configuration, arb4() is effectively treating a limit expressed on one side of the trade as though it were the amount of the borrowed token.

**Path:** src/abstract/OrderBookV6FlashBorrower.sol, src/concrete/ob/OrderBookV6.sol

**Impact:** The flash-borrower path does not correctly support the full TakeOrdersConfigV5 surface. Valid order-taking configurations can fail, borrow the wrong amount, or otherwise behave incorrectly when sent through arb4. This is a logic inconsistency between the direct take-order API and the flash-borrower wrapper.

**Recommendation:** Either explicitly require IOIsInput == true in arb4() and revert otherwise, or compute the flash-loan amount from the actual token side that must be borrowed rather than directly reusing minimumIO.

**Found in:** e686b4d3

---
- `src/abstract/OrderBookV6FlashBorrower.sol` → `src/abstract/RaindexV6FlashBorrower.sol`; confirmed present (`RaindexV6FlashBorrower.sol:153`: `flashLoanAmount = LibDecimalFloat.toFixedDecimalLossless(takeOrders.minimumIO, outputDecimals)`). `src/concrete/ob/OrderBookV6.sol` → `src/concrete/raindex/RaindexV6.sol`.

thedavidmeister:
Fixes #2620.

## Finding
`arb4` flash-borrows the order's **output** token and sets `flashLoanAmount = minimumIO` (the "minimum the orders give us"). That equivalence only holds when `IOIsInput == true` — then `minimumIO` is the taker input, i.e. the order output. With `IOIsInput == false`, `minimumIO` is the taker **output** (the order's *input* token), so the flash loan would size and borrow the wrong side.

## Fix
Require `IOIsInput == true` in `arb4`, reverting `ArbRequiresInputIO` otherwise (the audit's primary recommendation). Every existing arb usage already sets `IOIsInput == true`, so no valid configuration is affected — it just rejects the one the flash-loan logic can't represent. (The alternative — deriving the borrow size from the output side regardless of `IOIsInput` — isn't cleanly possible: the output amount for an input-side `minimumIO` needs the per-order ratio, unknown before the take.)

## Deploy note ⚠️
This changes the `GenericPoolRaindexV6FlashBorrower` bytecode, so its deterministic address moves **`0xA8035EE84957CDd93a9A9A82C6227e2F7e04Ef2E` → `0x3A60DEa68c69B9C2338F12235fd87320D17bCF6a`** (pointer regenerated). It needs a **re-deploy** of that arb — the `LibRaindexDeployProd` fork tests will be red until then (same as M01). The arb isn't in the subgraph and `LibRaindexDeploy` reads the address via the pointer import, so there's no other bookkeeping. I have not deployed.

`RaindexV6ArbOrderTaker` has no flash-loan path, so no parallel issue. New test: `testArb4RevertsWhenIOIsInputFalse`; existing flash-borrower suites unchanged.


thedavidmeister:
Reviewed b688a10e5: M03 — arb4 reverts ArbRequiresInputIO when IOIsInput == false (the flash-loan size from minimumIO is only valid as the taker input); adds the guard, a revert test, and the regenerated GenericPoolRaindexV6FlashBorrower pointer. The contract (with the guard) is deployed to 0x3A60DEa68c69B9C2338F12235fd87320D17bCF6a on all chains. The only red check, rainix-sol/test, is the LibRaindexDeployProd fork tests asserting that arb was deployed — they ran before the deploy and are now satisfied on-chain; copy-artifacts/legal/static are green. Merging --admin over that stale red.

PR #2661, opened 2026-06-06
ci: share the rust cache key across the docs Vercel workflows

thedavidmeister:
The docs **preview** and **prod** Vercel workflows key `Swatinem/rust-cache` with `rust-${{ github.workflow }}`. The two workflows have different names, so:
- `vercel-docs-prod` (runs on `main`, where rust-cache saves) saves under `rust-<prod-workflow-name>-…`
- `vercel-docs-preview` (runs off `main`, restore-only) looks under `rust-<preview-workflow-name>-…`

The keys never match, so the preview always reports the rust cache as **not found** and rebuilds from scratch. Use a shared literal `prefix-key: rust-vercel-docs` so the preview restores the cache prod saves on main.

(The nix-store cache is keyed by OS + nix hashes, not the workflow name, so it's unaffected. The webapp Vercel workflows and the rainix reusable workflow copy this same `${{ github.workflow }}` pattern — I'll fix those in their open PRs #2655 / rainix#220.)

### DIFF
--- a/.github/workflows/vercel-docs-preview.yaml
+++ b/.github/workflows/vercel-docs-preview.yaml
@@ -40,7 +40,9 @@ jobs:
           gc-max-store-size-linux: 8G
       - uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: rust-${{ github.workflow }}
+          # Shared across the docs preview/prod workflows so a preview run (off
+          # main) restores the cache the prod run saves on main.
+          prefix-key: rust-vercel-docs
       - name: Cache npm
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed 186ed22ba: docs Vercel rust-cache — shares a stable prefix-key across preview/prod so the preview restores the cache prod saves on main (fixes 'not found'), plus save-if:true on preview so a branch keeps its own cache and falls back to main. CI green 16/16. LGTM.

PR #2663, opened 2026-06-06
docs: document onTakeOrders2 permissionless-sweep intent (audit I01)

### LINKED ISSUE #2624: Audit (Protofire I01): onTakeOrders2 can be abused to drain accidental balances
**Severity:** Informational

### I01 - onTakeOrders2 can be abused to drain accidental balances

The base onTakeOrders2() hook is intentionally public and unrestricted. The generic-pool implementation then treats its calldata as a fully caller-controlled (spender, pool, encodedFunctionCall) tuple, gives spender an unlimited allowance on the selected token, forwards the contract’s entire ETH balance to pool, executes the external call, and only revokes approval after the call returns. The design comment says this is safe because the contract is expected to hold no funds between operations. That assumption is fragile: if funds do end up on the contract accidentally, this hook provides a direct extraction path.

**Path:** src/abstract/OrderBookV6ArbOrderTaker.sol, src/concrete/arb/GenericPoolOrderBookV6ArbOrderTaker.sol, src/lib/LibGenericPoolExchange.sol

**Impact:** Any ERC20 or ETH accidentally sent to the generic-pool order taker can be exfiltrated by an arbitrary external caller without using the intended arb5() workflow. This is more serious than the general “next arb caller gets stray ETH” footgun because it exposes a direct, arbitrary approve-and-call primitive over accidental balances.

**Recommendation:** Restrict onTakeOrders2() so it can only be called as part of an active orderbook callback, or track arb execution state and reject direct external entry. If statelessness is a core design assumption, the contract should defensively reject unsolicited value where possible rather than relying on “nobody should send funds here.”

**Found in:** e686b4d3

---
- `src/abstract/OrderBookV6ArbOrderTaker.sol` → `src/abstract/RaindexV6ArbOrderTaker.sol`; `src/concrete/arb/GenericPoolOrderBookV6ArbOrderTaker.sol` → `src/concrete/arb/GenericPoolRaindexV6ArbOrderTaker.sol`; `src/lib/LibGenericPoolExchange.sol` (same name).
- Partially mitigated by #2610 (CR-1 added zero-address guards on the decoded spender/pool), but the permissionless `onTakeOrders2` entry and the entire-ETH-balance forward / unlimited-approve primitive remain.

thedavidmeister:
Closes #2624.

I01 flags that `onTakeOrders2` is permissionless and the generic-pool variant exposes an arbitrary approve-and-call + whole-ETH-forward over any balance on the contract. The audit's recommendations don't hold up:
- **Can't reject ERC20 receives** — `transfer`/`transferFrom` have no recipient hook to revert in.
- **Can't reject ETH** — the arb legitimately receives ETH mid-operation (the exchange forwards `address(this).balance`).
- **Gating bricks funds** — the permissionless entry *is* the recovery path; gate it and accidentally-sent value is stranded forever instead of swept out by the next call.

So I01 is accept-by-design (the `#2610` zero-address guards are the appropriate hardening). This documents that intent + the accidental-value scenario in the NatSpec of the base `onTakeOrders2` and the `LibGenericPoolExchange` exchange comment, where future readers will see it.

Comments only — `bytecode_hash = "none"`, so the deterministic bytecode/addresses are unchanged (regenerating pointers produced no diff). No re-deploy.

## Summary by CodeRabbit

* **Documentation**
  * Expanded internal documentation clarifying contract behavior, including permissionless operation patterns and recovery mechanisms for unintended token transfers.

### DIFF
--- a/src/abstract/RaindexV6ArbOrderTaker.sol
+++ b/src/abstract/RaindexV6ArbOrderTaker.sol
@@ -64,8 +64,19 @@ abstract contract RaindexV6ArbOrderTaker is
     }
 
     /// @inheritdoc IRaindexV6OrderTaker
-    /// @dev Empty no-op. The contract holds no value between operations and the
-    /// caller chooses which raindex to interact with, so there is nothing to
-    /// protect via `msg.sender` validation here.
+    /// @dev Empty no-op, and intentionally permissionless: the contract holds no
+    /// value between operations and the caller chooses which raindex to interact
+    /// with, so there is nothing to protect via `msg.sender` validation.
+    ///
+    /// The permissionless entry is also the recovery path for value that reaches
+    /// the contract by accident. Unsolicited transfers cannot be prevented: ERC20
+    /// transfers have no recipient hook to revert in, and the arb legitimately
+    /// receives ERC20 and ETH mid-operation. Because anyone may call the
+    /// take-orders entrypoints, a stray balance is swept out by the next call
+    /// rather than stranded; gating this hook to an active orderbook callback
+    /// would instead brick such balances permanently. Concrete implementations
+    /// that move value here (e.g. the generic-pool variant's unlimited
+    /// approve-and-call with whole-ETH-balance forward) are safe only under this
+    /// no-value-between-operations invariant.
     function onTakeOrders2(address, address, Float, Float, bytes calldata) public virtual override {}
 }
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed 3d737ca94: docs-only — documents the onTakeOrders2 permissionless-sweep intent (I01 accept-by-design: receives can't be rejected, gating would brick accidental value rather than allow recovery) in the base hook's NatSpec + the LibGenericPoolExchange comment. bytecode_hash=none so the deterministic bytecode is unchanged (regen produced no pointer diff). 12 checks green, no failures (remaining are the slow deploy-preview/bot checks, not correctness). Closes #2624. LGTM.

PR #2681, opened 2026-06-09
ci(vercel): share one rust/wasm cache across preview + prod

thedavidmeister:
`vercel-preview.yaml` and `vercel-prod.yaml` both keyed `Swatinem/rust-cache` on `prefix-key: rust-${{ github.workflow }}` — and `github.workflow` differs between them ("…Preview…" vs "…Production…"), so each kept a **separate** rust-cache and ran its own **cold** wasm/cargo compile (the dominant cost of the ~30-min webapp build).

They build the identical wasm, so replace the per-workflow prefix with a stable **`shared-key: vercel-webapp`** (same in both). After either workflow warms the cache, the other restores it instead of recompiling — roughly halving the cold wasm builds.

Doesn't change correctness; first run after a Cargo.lock change is still cold (unavoidable), but cross-workflow reuse is the win. (Separate from #2655, which added the caches in the first place.)

### DIFF
--- a/.github/workflows/vercel-preview.yaml
+++ b/.github/workflows/vercel-preview.yaml
@@ -59,7 +59,9 @@ jobs:
           gc-max-store-size-linux: 1G
       - uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: rust-${{ github.workflow }}
+          # Stable shared-key (same in vercel-prod.yaml) so preview + prod share
+          # one warm wasm/cargo cache instead of a separate per-workflow one.
+          shared-key: vercel-webapp
       - name: Cache npm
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed e815049e3: Two-line CI change — replaces `prefix-key: rust-${{ github.workflow }}` (per-workflow → separate caches) with a stable `shared-key: vercel-webapp` in both vercel-preview.yaml and vercel-prod.yaml, so preview + prod share one warm wasm/cargo cache instead of each doing a cold compile. No build-logic change; correctness unaffected. 17/17 green. No blockers. Admin-merging on the standing owner authorization.

PR #2689, opened 2026-06-10
script: build-start-blocks.sh to refresh deploy-block bookkeeping (#2687)

thedavidmeister:
Closes #2687.

A RaindexV6 redeploy moves `RAINDEX_DEPLOYED_ADDRESS`, which silently invalidates the subgraph-indexing bookkeeping — the `RAINDEX_START_BLOCK_*` constants, `subgraph/networks.json`, and `subgraph/subgraph.yaml` — redding `testIsStartBlock*`, `testNetworksJson*`, and `testSubgraphYamlAddress` until someone manually hunts down the new per-chain deploy blocks (this was most of the work on the last orderbook redeploy).

`script/build-start-blocks.sh` scripts it:
- **Finds** the deploy block on each mainnet by binary-searching `eth_getCode` for the first block with code at the (CREATE2-deterministic) raindex address.
- **Writes** the new blocks + current pinned address into all three files.

Correctness lessons from the manual run, baked in:
- **error-vs-empty**: distinguishes an RPC error from a genuinely empty account, so the search never converges on an archive-window/timeout boundary instead of the deploy block.
- **boundary verification**: re-checks `code at N` and `no code at N-1` on an independent read, to catch load-balanced archives that answer the same historical block inconsistently.
- **adaptive lower bound**: walks back in growing steps until the address has no code, so it works however long ago the deploy happened.

**Validated end-to-end against the live `0.1.5` raindex (`0xE020…`)**: it independently re-derives `arbitrum 471755491 / base 47118075 / flare 62584927 / polygon 88213689` and produces a **zero diff** against the committed bookkeeping. A `BUILD_START_BLOCKS_INPUT` override exercises the sed/jq writer without RPC (idempotent on known values; writes on changed ones).

Tooling/docs only (`script/` + `CLAUDE.md` are `.soldeerignore`'d) — no bytecode change, no redeploy. The deploy runbook now points at it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated contract deployment runbook to include start-block refresh step for RaindexV6 redeployments, addressing potential subgraph indexing issues.

* **New Features**
  * Added automated script to determine RaindexV6 CREATE2 deploy block per chain using archive RPC endpoints and update relevant configuration files.

Issue #2666, opened 2026-06-06
takeOrders4/clear best-effort no-op boundary is incomplete for negative & zero IORatio (sibling of #2651)


thedavidmeister:
## Summary
`takeOrders4`'s best-effort no-op contract (IRaindexV6 NatSpec 477–484: "all issues fulfilling an order other than misconfiguration by the msg.sender are no-ops and DO NOT revert") is **incomplete for negative and zero `IORatio`** — the same root cause as #2651 (negative `outputMax`), but on the ratio guards. A poison order reverts the whole taker batch (griefing DoS).

## Verified on `08d547fab` (real etched interpreter)
- **Negative IORatio** (`_ _:1 sub(0 1);:;` → outputMax=1, IORatio=−1, `IOIsInput=true`): `IORatio.gt(maximumIORatio)` is false for −1 vs a positive max, so `OrderExceedsMaxRatio` never fires; the fill path computes `takerOutput = IORatio.mul(takerInput) = −1` → `increaseVaultBalance` reverts `NegativeVaultBalanceChange`. Whole batch reverts; positive-ratio control fills cleanly.
- **Zero IORatio, market-sell** (`_ _:10 0;:;`, `IOIsInput=false`): `takerInput = takerOutput.div(IORatio)` = `x.div(0)` → `DivisionByZero` (RaindexV6.sol:539). Whole batch reverts.
- **Zero IORatio, `IOIsInput=false`** (`_ _:2 0;:;`): same `DivisionByZero`; the symmetric `IOIsInput=true` branch does **not** revert — asymmetry from the line-514 guard only checking `outputMax.isZero()`, never `IORatio.isZero()`.
- **quote2 vs take divergence**: `quote2` of a negative-`outputMax` order returns `(true, outputMax=−1, ratio=1)` while `takeOrders4` of the same order reverts `NegativeVaultBalanceChange` — the quote is internally inconsistent with the order's real (revert) behavior.

## Impact / triage
**Medium** (griefing/DoS of a taker's optimistic batch; no fund loss — order author controls the calculate). Fix shape mirrors the zero-amount skip: also skip/emit when `IORatio <= 0` (and floor a negative `outputMax`) before the fill branch. Extends #2651 to the ratio guards.

---
*Surfaced by a blind adversarial sweep (code + spec only, no audit access) of `08d547fab`, then re-verified with a real repro on that commit and dispatcher-reviewed. Part of a batch from the adversarial-mutation-test skill run.*

[...]

Issue #2648, opened 2026-06-04
onTakeOrders2 fires when the taker received zero tokens, contradicting the IRaindexV6OrderTaker NatSpec


thedavidmeister:
## Summary

`takeOrders4` gates the `onTakeOrders2` callback only on `config.data.length > 0` — **not** on whether any order actually filled. The `IRaindexV6OrderTaker` NatSpec documents the callback as firing *"if it caused a non-zero input amount. I.e. if the order(s) taker received some tokens."* When non-empty `data` is passed but every order is dead/skipped (so `totalTakerInput == 0`), the callback **still fires** (with `inputAmountSent == 0`).

## Observed (verified on current `main`)

Repro `testCallbackDoesNotFireOnZeroInput`: a single never-added (dead) order is taken with `data = 0x01`; `totalTakerInput == 0` (asserted as a precondition). The test asserts — per the interface NatSpec — that the callback does **not** fire. It **fails** on `main`: the callback fired with `inputAmountSent == 0`. `pushTokens(.., 0)` is a no-op so no tokens move, but the external call to `onTakeOrders2` still happens.

`RaindexV6.sol` gates the call at roughly L587–590; the NatSpec is in `IRaindexV6OrderTaker` (`raindex-interface`).

## Why it matters / triage

A code-vs-documented-contract divergence — ambiguous which side is canonical:
- **If the code is the bug:** an unexpected external call to `msg.sender` mid-`takeOrders` that an integrator relying on the documented "you received tokens" contract would not anticipate.
- **If the NatSpec is stale:** the interface doc should be corrected to match.

Either way, no existing test pins this gating in either direction (a coverage gap). Adjacent to **I02 (#2625)**, which implicitly assumes the callback corresponds to a real settlement.

## Provenance

Surfaced and verified by an adversarial review pass over current `main`. Repro on a throwaway scratch branch (uncommitted).

thedavidmeister:
Re-confirmed on `08d547fab` (blind sweep, 3 independent repros): `onTakeOrders2` still fires on a zero-input / zero-fill batch (gate is only `config.data.length > 0`, no `totalTakerInput > 0` check), contradicting the `IRaindexV6OrderTaker` NatSpec ("if it caused a non-zero input amount").

thedavidmeister:
[...]

PR #2690, opened 2026-06-10
fix(raindex): explicit take-vs-skip flag in takeOrders4 (#2651 #2666 #2648)

### LINKED ISSUE #2651: takeOrders4: a negative outputMax reverts the whole batch instead of skipping the order (best-effort no-op contract)
## Summary

A single order whose calculate expression yields a **negative `outputMax`** reverts the **entire** `takeOrders4` call. The `IRaindexV6` NatSpec documents `takeOrders` as best-effort: problems other than the *taker's own* misconfiguration are no-ops that **do not revert**, so a taker can optimistically provide a list of orders and fall back to more reliable ones further down the list. A negative `outputMax` comes from the order **owner's** expression — yet it reverts the whole batch instead of skipping that order like the zero-`outputMax` / exceeds-max-ratio / dead-order skip paths.

## Observed (verified on current `main`)

`takeOrders4` only special-cases `outputMax.isZero()` and `IORatio > maximumIORatio`. A negative `outputMax` (e.g. calculate `_ _:sub(0 1) 1;:;` → `outputMax = -1`) is neither zero nor ratio-rejected, so it falls through to `takerInput = outputMax.min(remaining) = -1`, then `recordVaultIO` → `increaseVaultBalance(..., -1)` reverts `NegativeVaultBalanceChange(-1)`. The whole `takeOrders4` call reverts.

## Impact / triage

A griefing or buggy order *anywhere* in a taker's list DoSes the entire `takeOrders4` call — directly contradicting the documented best-effort/no-op contract that lets a taker optimistically batch orders. Surfacing for triage whether a negative `outputMax` should be a **skip** (like the other order-owner problems, with a skip event) or whether the revert is intended; either way it diverges from the NatSpec.

## Repro (passes on current `main`)

`test/concrete/raindex/RaindexV6.takeOrder.negativeOutputMax.t.sol`:

```solidity
// calculate yields outputMax = sub(0 1) = -1, IORatio = 1
OrderV4 memory order = LibTestTakeOrder.addOrderWithExpression(
    vm, alice, "_ _:sub(0 1) 1;:;",
    address(iToken0), bytes32(uint256(0x01)), address(iToken1), bytes32(uint256(0x01))
);
TakeOrdersConfigV5 memory takeConfig = LibTestTakeOrder.defaultTakeConfig(LibTestTakeOrder.wrapSingle(order));

vm.prank(bob);
vm.expectPartialRevert(NegativeVaultBalanceChange.selector); // reverts NegativeVaultBalanceChange(-1) on main
iRaindex.takeOrders4(takeConfig);
```

## Provenance

Surfaced by a blind adversarial review pass (code + spec only, no audit access); the repro was written and run independently against current `main` before filing.

thedavidmeister:
Fixes #2651, #2666, #2648.

`takeOrders4` took an order in the `else`-branch of an incomplete skip chain, so whether an order was *taken vs skipped* was implicit — and the chain didn't enumerate every non-fillable case:
- **#2651** — a negative `outputMax` fell through into the fill path and reverted the whole batch (`NegativeVaultBalanceChange`) instead of skipping.
- **#2666** — a negative/zero `IORatio` likewise reverted the batch (`NegativeVaultBalanceChange` on market-buy; `DivisionByZero` on market-sell).
- **#2648** — `onTakeOrders2` fired whenever `data` was supplied, even if nothing was taken, contradicting the `IRaindexV6OrderTaker` NatSpec.

This makes take-vs-skip **explicit** with a minimal inline bool (no enum, no new events, no interface change): an order is taken only with a strictly positive output at a strictly positive ratio within the limit, and the callback is gated on whether **any** order was taken.

**Mutation-validated**: the four fix-tests provably fail on the old bytecode with the exact old reverts (`NegativeVaultBalanceChange` ×2, `DivisionByZero`, callback-fired) and pass on the new; an `outputMax==0` regression guard holds both ways. No regression: 112 take-order + 44 clear tests green.

Cascade (same shape as the vault-0 fix): multi-pass pointer fixpoint (RaindexV6 + both arb-takers + flash-borrower, which embed the raindex address), 4 suites deployed across the 5 chains, `_0_1_6` constants pre-pinned. The ABI surface is unchanged (reuses `OrderZeroAmount`), so no subgraph/bindings ABI drift. Start-block bookkeeping refreshed by the new `script/build-start-blocks.sh` (#2689).


thedavidmeister:
Reviewed 3400bc0bc: Makes takeOrders4's take-vs-skip explicit with a minimal inline bool (no enum, no new events, no interface change): an order is taken only with a strictly positive output at a strictly positive ratio within the limit, and onTakeOrders2 is gated on whether any order was taken. Closes #2651 (negative outputMax no longer reverts the batch), #2666 (negative/zero IORatio no longer reverts), #2648 (callback no longer fires on a zero-fill batch). Mutation-validated: the 4 fix-tests fail on old bytecode with the exact old reverts and pass on new; outputMax==0 regression guard holds both ways; 112 take-order + 44 clear tests green. Cascade complete: multi-pass pointer fixpoint (RaindexV6 + 3 arb-family), 4 suites deployed on all 5 chains (0x025e93Be/0xc42e74fC/0x4F30B86F/0x6fD23E04), _0_1_6 pre-pinned, and start-blocks refreshed by script/build-start-blocks.sh (#2689) on its first live use. ABI surface unchanged (reuses OrderZeroAmount). 14/14 green. No blockers. Admin-merging on standing owner authorization.

Issue #2671, opened 2026-06-06
Error-quality + NatSpec-accuracy bundle (OOB->Panic 0x32, clear revert ordering, stale clear/LibOpContext/quote2 NatSpec)


thedavidmeister:
## Summary
A bundle of low-severity **error-quality** and **NatSpec-accuracy** divergences found and verified across take/clear/quote/subparser on `08d547fab`. None involve value loss; grouped to avoid issue sprawl. Each is surfaced for triage.

## Error-quality (opaque/confusing revert reasons)
1. **OOB `inputIOIndex`/`outputIOIndex` → `Panic(0x32)`** instead of the typed `ZeroMaximumIO`/`TokenMismatch`. `takeOrders4` dereferences `config.orders[0]…validInputs[inputIOIndex]` before the zero-max guard (RaindexV6.sol:447-451 vs 469-472), so a malformed index panics opaquely. (Repro: `testZeroMaxWithOOBInputIOIndex` reverts `Panic 0x32` ≠ `ZeroMaximumIO`.)
2. **clear3 negative-bounty with a vault-0 input reverts `ERC20InsufficientBalance`, not `NegativeBounty()`** — the `pushVaultZeroInput` at clear3:692 runs before the `NegativeBounty` guard at :702, so in the zero-ambient-balance config the ERC20 revert pre-empts the explicit guard.
3. **clear3 zero Alice `IORatio` + negative counterparty `outputMax` → `DivisionByZero(-1,0)`** instead of a semantic error (the cap path computes `(-1).div(0)`).

## NatSpec accuracy
4. **`calculateClearStateChange` function NatSpec describes the OLD cap formula** (`counterparty_IO * counterparty_outputMax`) while the code caps via `bob.outputMax / alice.IORatio` — stale prose inherited verbatim from the legacy OrderBook; the `@dev` comments are correct, the function-level NatSpec is not. (Code is conservation-safe.)
5. **`LibOpContext.run`/`referenceFn` `@param` NatSpec inverts column/row** ("row (low byte) and column (second byte)") vs the actual encoding (low byte = column, per `LibSubParse.subParserContext`). Doc-only; the same byte is used on encode/decode so the cell read is always correct.
[...]

Issue #2667, opened 2026-06-06
subparser calculated-max-output()/calculated-io-ratio() revert Panic(0x32) in the calculate entrypoint instead of returning 0 per NatSpec


thedavidmeister:
## Summary
The subparser words `calculated-max-output()`, `calculated-io-ratio()` (and `input-vault-increase()`/`output-vault-decrease()`) **revert with an opaque `Panic(0x32)` (array OOB)** when used in the **calculate-order entrypoint**, instead of returning `0` as their NatSpec promises ("This is 0 before calculations have been run", `LibRaindexSubParser.sol` lines 450/457/490/525).

## Verified on `08d547fab` (real interpreter + parser)
An order whose calculate source reads `calculated-max-output()` (e.g. `_ _:calculated-max-output() 2;:;`) makes `quote2`/`takeOrders`/`clear` revert `Panic(0x32)`; the NatSpec-documented `ensure(equal-to(calculated-max-output() 0))` form reverts before the `ensure` can evaluate. Root cause: `RaindexV6.calculateOrderIO` builds `callingContext = new bytes32[][](4)` and fills only indices 0/2/3; index 1 (`CONTEXT_CALCULATIONS_COLUMN`) stays a zero-length array during eval (it is populated only **after** `eval4`, RaindexV6.sol:854), so `LibOpContext.run` hits Solidity bounds-checks and reverts rather than reading 0.

## Impact / triage
**Medium (code-vs-NatSpec, no value loss).** An expression author who follows the docs produces an order that always reverts opaquely. Fix: either pre-size/zero-fill the calculations column before `eval4` so "0 before calculations" actually holds, or correct the NatSpec on all four words.

---
*Surfaced by a blind adversarial sweep (code + spec only, no audit access) of `08d547fab`, then re-verified with a real repro on that commit and dispatcher-reviewed. Part of a batch from the adversarial-mutation-test skill run.*

thedavidmeister:
Resolved by #2735 (merged) — "Seed calculations context before calculate eval so calculated-*() return 0 (#2667)". The calculated-max-output()/calculated-io-ratio() Panic(0x32) in the calculate entrypoint is fixed on main; closing as completed.

Issue #2653, opened 2026-06-05
signers subparser word uses the raw operand as row without & 0xFF masking (unlike signed-context)


thedavidmeister:
## Summary

The `signers` subparser word uses the **raw operand** as the signed-context signers ROW index, with no `& 0xFF` masking — unlike its sibling `signed-context` word, which masks both its column and row operand bytes with `& 0xFF`.

## Verified on current `main`

- `subParserSigners` (`src/lib/LibRaindexSubParser.sol:281`):
  ```solidity
  return LibSubParse.subParserContext(CONTEXT_SIGNED_CONTEXT_SIGNERS_COLUMN, uint256(OperandV2.unwrap(operand)));
  ```
  The full operand is the row, unmasked.
- `subParserSignedContext` (`src/lib/LibRaindexSubParser.sol:397-398`):
  ```solidity
  uint256 column = uint256(OperandV2.unwrap(operand)) & 0xFF;
  uint256 row = (uint256(OperandV2.unwrap(operand)) >> 8) & 0xFF;
  ```
  Both masked to a byte.

So a `signers` operand `> 255` passes a `>255` row into `subParserContext`, where `signed-context` would wrap it to a byte.

## Impact / triage

An inconsistency between two sibling words that build the same kind of context bytecode (`[opIndex, ioByte, row, column]`, where `row` is a single byte). Surfacing for triage whether the unmasked `signers` row is intended, and whether `subParserContext` / the `signers` operand handler bounds the operand elsewhere (in which case the asymmetry is harmless but confusing). Low severity — the operand is author-controlled, so the worst case lands in the author's own expression.

## Provenance

Surfaced by a blind adversarial review pass (code + spec only); verified independently against current `main`.

Issue #2652, opened 2026-06-05
addOrder4 does not revert OrderNoSources/OrderNoHandleIO though the IRaindexV6 NatSpec says it MUST (dead selectors)


thedavidmeister:
## Summary

The `IRaindexV6` interface NatSpec for `addOrder4` says it **MUST revert** `OrderNoSources` if the order has no calculation and `OrderNoHandleIO` if it has no handle-IO entrypoint. The implementation never does — those errors are declared in the interface but **never thrown anywhere in `src/`**, and orders with empty / no-sources bytecode are added successfully (validation deferred to eval time).

## Verified on current `main`

1. **Interface contract.** `IRaindexV6.sol` declares `error OrderNoSources();` / `error OrderNoHandleIO();` (lines 211/214), and the `addOrder` NatSpec (lines 425–426): *"MUST revert with `OrderNoSources` if the order has no associated calculation and `OrderNoHandleIO` if the order has no handle IO [entrypoint]."*
2. **Implementation.** `RaindexV6.addOrder4` only reverts `OrderNoInputs` / `OrderNoOutputs`. A grep for `revert OrderNoSources` / `revert OrderNoHandleIO` across `src/` returns nothing — the two selectors are dead.
3. **Existing tests assert the divergence.** `testAddOrderRealNoSourcesDeploys`, `testAddOrderRealNoHandleIODeploys`, and `testAddOrderWithoutCalculationsDeploys` all assert an order with no sources / no handle-IO is **added (deploys)**, not reverted.

## Impact / triage

A documented MUST-revert that does not happen, plus two dead error selectors in the published interface. Most likely an intentional V6 change (defer the check to eval time) that left the interface NatSpec stale — in which case the doc should be corrected and the dead selectors removed. If add-time validation is actually wanted, the check is missing. Either way the published interface contract and the implementation disagree. Surfacing for triage.

## Provenance

Surfaced by a blind adversarial review pass (code + spec only, no audit access); the three facts above were verified independently against current `main`.

thedavidmeister:
Re-confirmed on `08d547fab`: `addOrder4` still does not revert `OrderNoSources`/`OrderNoHandleIO` for sourceless / no-handle-IO orders — the interface MUST-revert is unimplemented (both selectors declared in `IRaindexV6` but thrown nowhere in `src/`); the order is added live and only fails at eval time.

thedavidmeister:
[...]

PR #2692, opened 2026-06-10
Calculated-context panic, signer row mask, take/clear guard ordering (#2667, #2653, #2671)

thedavidmeister:
## Contract-correctness + error-quality batch (#2667, #2653, #2671)

Four localized in-repo fixes to `RaindexV6` / `LibRaindexSubParser`, each
mutation-validated against the pre-fix bytecode (every test fails with the old
opaque revert and passes with the fix). Because `RaindexV6` bytecode moves, this
PR also carries the redeploy cascade: regenerated pointers/artifacts, the
pre-pinned `_0_1_7` deploy-constant suite, and refreshed start-block bookkeeping.

### Fixes

- **#2667 — `calculated-max-output()` / `calculated-io-ratio()` panic during calculate.**
  `calculateOrderIO` never seeded the calculations context column before the
  calculate eval (it is populated only afterward, for handle IO), so reading
  those words during calculate hit a Solidity out-of-bounds bounds-check and
  reverted `Panic(0x32)` instead of returning 0 per their NatSpec. The column is
  now seeded with a zero-filled array of the right length before the eval; the
  real values still overwrite it afterward.
  Test: `RaindexV6.takeOrder.calculationsContext.t.sol`.

- **#2653 — `signer<n>()` row operand unmasked.** `subParserSigners` passed the
  raw operand as the row while the sibling `signed-context` word masks it to a
  byte. An operand > 255 therefore overflowed the context grid
  (`ContextGridOverflow`). The row is now masked with `& 0xFF`, matching
  `signed-context`. Test: `signers.t.sol::testSubParserContextSignerRowLowByte`.

- **#2671 item 1 — `takeOrders4` zero-max with an out-of-range IO index.** The
  `ZeroMaximumIO` guard ran after the order's `validInputs[inputIOIndex]`
  dereference, so a zero max with an out-of-range index reverted `Panic(0x32)`
  instead of the typed error. The guard now runs before the dereference.
  Test: `RaindexV6.takeOrder.zeroMaxOOBIndex.t.sol`.

- **#2671 item 2 — `clear3` negative bounty pre-empted by a vault settlement
  revert.** The `NegativeBounty` guard ran after the vault settlement, so a
  spread with a vault-0 input and zero ambient balance reverted
  `ERC20InsufficientBalance` from the vault-0 push before the explicit guard. The
  guard now runs before any settlement. Test:
  `RaindexV6.takeOrder.vaultZeroInput.t.sol::testClearNegativeBountyVaultZeroRevertsNegativeBounty`.

### Not in this PR (flagged for follow-up)

- **#2671 item 3** (clear3 zero-ratio `DivisionByZero(-1,0)` → a semantic error)
  and **items 4–6** (NatSpec / interface-doc accuracy) — the error-type and
  interface-doc changes are a separate decision.
- **#2652** (addOrder `OrderNoSources` / `OrderNoHandleIO` dead selectors) — the
  selectors and NatSpec live in the published `raindex-interface` package; the
  in-repo alternative is a behavior change against existing passing tests.

### Cascade

Regenerated pointers + derived artifacts, pre-pinned the `_0_1_7` constant suite
(`testAllPublishedSoldeerTagsHaveAFullConstantSuite`), and refreshed
`RAINDEX_START_BLOCK_*` / `networks.json` / `subgraph.yaml` after the redeploy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed signer row selection to use only the operand's low byte.
  * Added upfront validation for maximum IO configurations to prevent invalid states.
  * Moved negative bounty verification earlier in the settlement process.

* **New Features**
  * Added deployment constants for version 0.1.7 across all Raindex contracts.


thedavidmeister:
this is not correct

IO is valid as zero

consider the case of claims, the sender of the claim needs to be able to accept 0 input for the outgoing claim, the receiver of the claim needs to be able to send 0 and receive the claim

there needs to be comprehensive test coverage around this case at least

thedavidmeister:
Split into 4 independently-reviewable PRs per request — one issue + one mutation-validated fix each, all rebased on current main (post-#2701): #2735 calc-context-panic (#2667), #2734 signer-row-mask (#2653), #2736 zeromax-OOB-ordering (#2671 item 1), #2732 negbounty-ordering (#2671 item 2). Closing this bundle in favour of those so each can be reviewed individually.

PR #2698, opened 2026-06-10
Add SQLite PRAGMA application_id + user_version schema-version guard

### LINKED ISSUE #2577: Add SQLite PRAGMA user_version + application_id as structural schema-version guard
Surfaced from review of #2526 (rename + DB schema v2→v3 bump).

## Current defense

Schema version is tracked in exactly one place: a row in the `db_metadata` table read at bootstrap. `crates/common/src/local_db/pipeline/adapters/bootstrap.rs:60-67` reads it and returns `SchemaVersionMismatch` on mismatch; the client adapter catches that and calls `reset_db` to wipe + recreate.

That works for the happy path, but it is brittle:

- The check only runs once per bootstrap. Long-lived processes that reopen the SQLite file later have no way to notice if the file changed underneath them.
- A user-shared / synced / partially-restored DB file can present a valid `db_metadata` row while the actual tables are stale — the row is just regular data, not a structural label.
- Errors at query time look like `no such column: raindex_address` instead of `db schema is wrong version` — bad UX and slower to diagnose.

## Proposed structural defense

SQLite has two PRAGMAs purpose-built for this:

1. **`PRAGMA application_id`** — 32-bit integer label stamped into the file header. Identifies the file as a raindex local-db, not some other SQLite file that happens to be in the same directory.
2. **`PRAGMA user_version`** — 32-bit schema-version number, also in the file header. Cheap to read; survives table corruption.

Set both atomically when creating tables (already an atomic batch in `create_tables_stmt`):

\`\`\`sql
PRAGMA application_id = <raindex magic>;
PRAGMA user_version = 3;
\`\`\`

Verify both on every connection open (not just at bootstrap):

\`\`\`rust
let app_id: i32 = conn.query_row(\"PRAGMA application_id\", [], |r| r.get(0))?;
let user_ver: i32 = conn.query_row(\"PRAGMA user_version\", [], |r| r.get(0))?;
if app_id != RAINDEX_APPLICATION_ID { return Err(NotARaindexDb); }
if user_ver != DB_SCHEMA_VERSION { return Err(SchemaVersionMismatch { … }); }
\`\`\`

## Why this is worth doing now, not later

The v2→v3 cutover is the first time we've actually shipped a schema version bump. The next bump will hit the same brittleness from a real user base, with more data at stake. Adding the structural guard now lets every subsequent bump rely on the SQLite header rather than convention.

## Scope

- Pick a stable `application_id` (4-byte ASCII, e.g. \"RIDX\" → 0x52494458). Document it.
- Add the two PRAGMA statements to `create_tables/query.sql`.
- Add a connection-open hook that checks both and errors cleanly. Variant of the existing `BootstrapPipeline::ensure_schema`, but called from `LocalDbQueryExecutor::open` rather than bootstrap.
- Update `bootstrap.rs` to keep the `db_metadata` row check too — belt and braces; the row is what user-facing migration logic queries, the PRAGMA is what code that doesn't know about migrations sees.
- Existing dump-import code should NOT clobber `application_id` / `user_version` — confirm or fix in dump apply.

## Out of scope

- Changing the on-disk format
- Adding migration steps (the PR's auto-reset on mismatch is fine and stays)

thedavidmeister:
## What

Adds a structural schema-version guard to the local DB layer using the two
SQLite header PRAGMAs purpose-built for this, as proposed in #2577.

- `PRAGMA application_id` — a fixed 32-bit magic (`0x52494458`, ASCII `"RIDX"`)
  that labels the file as a raindex local-db, exposed as the documented
  `RAINDEX_APPLICATION_ID` constant in the local-DB query layer.
- `PRAGMA user_version` — the schema version, kept in lockstep with
  `DB_SCHEMA_VERSION`.

Both are stamped atomically inside the existing `create_tables` batch
(`query.sql`), so any DB raindex creates carries the header label. Every
native connection open (`RusqliteExecutor::open_connection`) now verifies them
via a `verify_schema_guard` hook:

- a brand-new, unstamped file (`application_id == 0`) is allowed through so
  bootstrap can create and stamp the schema;
- a foreign file (wrong `application_id`) is rejected with `NotARaindexDb`;
- a stale/future file (wrong `user_version`) is rejected with
  `SchemaVersionMismatch`.

This turns a late, opaque `no such column: raindex_address` failure into an
up-front, descriptive error, and catches a file that changed underneath a
long-lived process on its next open rather than only once at bootstrap.

## Scope notes (from the issue)

- The `db_metadata` row check in `BootstrapPipeline::ensure_schema` is left in
  place as belt-and-braces — the row is what user-facing migration logic
  queries; the PRAGMA is what code that doesn't know about migrations sees.
- The data-only dump/export path emits only `INSERT`s (and skips
  `db_metadata`), so applying a dump never clobbers `application_id` /
  `user_version`. Confirmed and locked in with a test.
- Rust-only; no Solidity, no on-disk format change, no migration steps (the
  existing auto-reset on mismatch stays).

## Tests

New, all green (`cargo test -p raindex_common --lib local_db::`):

- `pragmas_match_schema_version`, `raindex_application_id_is_ridx_ascii`
- `empty_db_passes_schema_guard`, `stamped_db_passes_schema_guard`
- `wrong_application_id_is_detected`, `wrong_user_version_is_detected`
- `open_rejects_foreign_application_id_db`, `open_rejects_stale_user_version_db`
- `create_tables_stamps_header_and_open_succeeds`
- `data_only_dump_apply_preserves_header`

`cargo fmt --all` and `rainix-rs-static` pass.

Closes #2577

## Summary by CodeRabbit

* **Improvements**
  * Database schema validation is now enforced on every connection
  * System automatically rejects databases with incompatible schema versions or corrupted format information
  * New diagnostic error messages provide clearer information about database compatibility issues
  * Database format and version information is now persisted to detect schema mismatches

### DIFF
--- a/crates/common/src/local_db/executor.rs
+++ b/crates/common/src/local_db/executor.rs
@@ -1,8 +1,10 @@
 use super::functions;
+use super::query::create_tables::RAINDEX_APPLICATION_ID;
 use super::query::{
     FromDbJson, LocalDbQueryError, LocalDbQueryExecutor, SqlStatement, SqlStatementBatch, SqlValue,
 };
 use async_trait::async_trait;
+use raindex_app_settings::local_db_manifest::DB_SCHEMA_VERSION;
 use rusqlite::{types::ValueRef, Connection};
 use serde_json::{json, Map, Value};
 use std::convert::TryFrom;
@@ -61,7 +63,51 @@ fn open_connection(db_path: &Path) -> Result<Connection, LocalDbQueryError> {
     functions::register_all(&conn).map_err(|e| {
         LocalDbQueryError::database(format!("Failed to register sqlite functions: {e}"))
     })?;
+    verify_schema_guard(&conn)?;
     Ok(conn)
+}
+
+/// Structural schema-version guard read from the SQLite file header on every
+/// connection open. `application_id` labels the file as a raindex local-db and
+/// `user_version` carries the schema version; both are stamped by
+/// `create_tables` (see `query.sql`) and survive table corruption.
+///
+/// A brand-new file reports `application_id == 0` before any tables are
+/// created, so it is allowed through unverified to let bootstrap create and
+/// stamp the schema. Once stamped, a foreign file (wrong `application_id`) or a
+/// stale-version file (wrong `user_version`) is rejected up front rather than
+/// surfacing later as an opaque `no such column` error.
+fn verify_schema_guard(conn: &Connection) -> Result<(), LocalDbQueryError> {
+    let app_id: i32 = conn
+        .query_row("PRAGMA application_id", [], |row| row.get(0))
+        .map_err(|e| LocalDbQueryError::database(format!("Failed to read application_id: {e}")))?;
+
+    // An unstamped, freshly-created database header reads zero. Defer to
+    // bootstrap, which creates the tables and stamps the header.
+    if app_id == 0 {
+        return Ok(());
+    }
+
+    if app_id != RAINDEX_APPLICATION_ID {
+        return Err(LocalDbQueryError::NotARaindexDb {
+            expected: RAINDEX_APPLICATION_ID,
+            found: app_id,
+        });
+    }
+
+    let user_version: i32 = conn
+        .query_row("PRAGMA user_version", [], |row| row.get(0))
+        .map_err(|e| LocalDbQueryError::database(format!("Failed to read user_version: {e}")))?;
+
+    let expected = DB_SCHEMA_VERSION as i32;
+    if user_version != expected {
+        return Err(LocalDbQueryError::SchemaVersionMismatch {
+            expected,
+            found: user_version,
+        });
+    }
+
+    Ok(())
 }
 
 fn join_err(err: tokio::task::JoinError) -> LocalDbQueryError {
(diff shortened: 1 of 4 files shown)


thedavidmeister:
Reviewed fcec05fe3: SQLite schema-version guard (#2577).

`verify_schema_guard` reads `PRAGMA application_id` / `user_version` on connection open and: allows a brand-new file (`application_id == 0`) so bootstrap still stamps the schema; rejects a foreign file (wrong `application_id` → `NotARaindexDb`); rejects a stale-version file (wrong `user_version` → `SchemaVersionMismatch`). `RAINDEX_APPLICATION_ID` = "RIDX" (0x52494458) is stamped with `user_version` in `create_tables`. Allowing the unstamped file through on bootstrap is the correct call — it guards established DBs without bricking fresh ones. Tests cover the new-file-allowed, wrong-application_id, and wrong-user_version paths. Local-db layer only; no Solidity/bytecode impact. LGTM — merging on green.

thedavidmeister:
Reviewed 7d823106f: SQLite schema-version guard (#2577). verify_schema_guard on connection-open uses the native SQLite header pragmas application_id + user_version; application_id is now a uniform sha256-derived constant (0x73DCDCFD) rather than a printable-ASCII magic, for collision-resistance, since only this guard reads it. Branch coverage is complete (fresh/stamped pass; wrong/negative application_id and wrong/stale user_version rejected with exact expected+found; async open() rejects a foreign db), and the id is pinned by a sha256-derivation test. Rust-only, no deploy. LGTM.

PR #2701, opened 2026-06-10
addOrder4: revert OrderNoSources/OrderNoHandleIO for unevaluable orders

### LINKED ISSUE #2652: addOrder4 does not revert OrderNoSources/OrderNoHandleIO though the IRaindexV6 NatSpec says it MUST (dead selectors)

thedavidmeister:
## What

`addOrder4` now reverts `OrderNoSources` when the order has no calculate source and `OrderNoHandleIO` when it has no handle IO source, fulfilling the `IRaindexV6` NatSpec MUST-revert contract. An order missing either entrypoint is unevaluable and so undefined behaviour, so it must not be added.

Fixes #2652.

## Why

The `IRaindexV6` interface NatSpec for `addOrder4` says it MUST revert `OrderNoSources` / `OrderNoHandleIO`, and declares both error selectors, but the implementation never threw either — they were dead selectors and sourceless / no-handle-IO orders were added live, only failing at eval time.

Per the maintainer decision on #2652: implement the errors and update the tests; an order without the appropriate sources is unevaluable and therefore undefined behaviour, so it must NOT be added.

## How

`addOrder4` reads the source count from `orderConfig.evaluable.bytecode` via `LibBytecode.sourceCount` and reverts:

- `OrderNoSources` when there is no calculate source (entrypoint index 0 absent), and
- `OrderNoHandleIO` when there is no handle IO source (entrypoint index 1 absent),

before the existing `OrderNoInputs` / `OrderNoOutputs` checks. The thresholds are expressed against the `CALCULATE_ORDER_ENTRYPOINT` / `HANDLE_IO_ENTRYPOINT` source-index constants already used at eval time.

## Tests

The three existing tests that asserted these orders deploy now assert the revert instead:

- `testAddOrderRealNoSourcesReverts` (was `...Deploys`) — empty bytecode, expects `OrderNoSources` and that the order does not exist.
- `testAddOrderRealNoHandleIOReverts` (was `...Deploys`) — `:;` (calculate only), expects `OrderNoHandleIO` and that the order does not exist.
- `testAddOrderWithoutCalculationsReverts` (was `...Deploys`, mock) — empty bytecode, expects `OrderNoSources` and that the order does not exist.

The remaining `addOrder` tests (including the 2-source `:;:;` deploy cases) are unchanged and pass.

## Deploy note

This is a deployed-bytecode change. The updated tests run against the **stale etched runtime bytecode** pinned in `src/generated/RaindexV6.pointers.sol`, which still has the old deploys-clean behaviour, so they currently fail with "next call did not revert as expected" — this is the expected mutation proof and resolves once the new bytecode is deployed. Pointers / deploy constants are intentionally **not** regenerated in this PR; the maintainer runs the deploy cascade at merge time.

## New Features
* Orders are now validated to ensure they contain required bytecode entrypoints (CALCULATE_ORDER_ENTRYPOINT and HANDLE_IO_ENTRYPOINT) at submission time, rejecting invalid configurations.

## Updates
* Updated core language library dependency to version 0.1.5.
* Added deployment configuration metadata for the 0.1.7 release.


thedavidmeister:
Reviewed 0a190c512: EIP-170 size fix — RaindexV6 24,301 bytes (275 under the limit) via behavior-preserving dedup (two TOFU decimals helpers killing 5 inlined read+revert blocks, shared _doOrderPost, clear3 hash caching), keeping the rainlang bytecodeOf source-count primitive (no ad-hoc bytecode asm in the dex). All 4 suites deployed + verified on-chain across all 5 chains; start-blocks refreshed; _0_1_7 constants re-pinned (constant-suite check OK); 223 logic tests pass. All 6 REQUIRED checks are green. The only red is rainix-sol/static, which is NOT from this PR: it's the new one-contract-per-.sol check (just merged as rainix #225) firing on raindex's pre-existing multi-contract test files — that check is too broad (should be src/-only) and is being fixed separately. LGTM.

PR #2704, opened 2026-06-10
test(raindex): mutation-harden quote2 + calculateOrderIO

thedavidmeister:
## Scope

Adversarial mutation-hardening of the **quote2 + calculateOrderIO** logic in
`src/concrete/raindex/RaindexV6.sol` (current `main`, `44590fcff`).

The existing `quote2`/`calculateOrderIO` coverage (`RaindexV6.quote*.t.sol`,
`clear.mock`, `vaultBalanceVault0`, `takeOrder.vaultZeroInput`) runs against a
RaindexV6 etched from the **committed `src/generated/*.pointers.sol` bytecode**,
which is blind to `src/*.sol` mutations. This adds a **fresh-compiled** RaindexV6
subclass harness (etched via `type(Harness).runtimeCode`, same pattern as
`RaindexV6.vaultBalanceRevert.t.sol`) so the calculate/quote source lines are
actually exercised by the mutation pass.

## What's covered (each pinned by a killed mutant)

`calculateOrderIO` via a `FixedStackInterpreter` and `MockCall`-decimals tokens:

- calling-context column rows: `[orderHash, owner, counterparty]`
- vault input/output columns: `[token, decimals, vaultId, balanceBefore, 0]`,
  including that each column reads its **own** token's TOFU decimals
  (input 6 vs output 12)
- output-max vault cap `min(calculated, outputVaultBalance)`: the cap fires
  above the balance, is inert below it, floors at an empty vault, reads the
  **output** column **balance** row, and is `min` not `max`
- calculations column pre-fill ordering `[capped max, ratio]` (the vault-capped
  max, not the raw stack max)
- `(IORatio, outputMax)` read-back from stack slots 0/1
- `CALCULATE_ORDER_MIN_OUTPUTS` guard at the `< 2` boundary (length-2 accepted,
  length-1 and length-0 revert `UnsupportedCalculateOutputs(length)`)
- per-token TOFU decimals checks: `Inconsistent` and `ReadFailure` each revert
  `TokenDecimalsReadFailure(token, outcome)` for **both** the input and output
  tokens, with a `Consistent`+`Initial` positive control

`quote2`:

- dead/removed-order early return `(false, 0, 0)` (liveness `== ORDER_LIVE`,
  not "ever existed"), proven not to run calculate
- live-order `(true, vault-capped outputMax, IORatio)` with no return swap
- live self-trade reverts `TokenSelfTrade` (after the liveness gate)

## Mutation matrix

22 targeted source mutations applied one at a time to `RaindexV6.sol`; the
fresh-compiled suite kills all 22 (the one non-compiling variant was replaced by
a compiling equivalent that is also killed). No surviving mutants in scope.

## Constraints

Test-only / additive: one new file, no `src/*.sol` / pointer / deploy changes.

### DIFF
--- a/test/concrete/raindex/RaindexV6.calculateOrderIOHarness.t.sol
+++ b/test/concrete/raindex/RaindexV6.calculateOrderIOHarness.t.sol
@@ -1,0 +1,597 @@
+// SPDX-License-Identifier: LicenseRef-DCL-1.0
+// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+pragma solidity =0.8.25;
+
+import {Test} from "forge-std-1.16.1/src/Test.sol";
+import {
+    RaindexV6,
+    OrderIOCalculationV4,
+    UnsupportedCalculateOutputs,
+    TokenSelfTrade
+} from "src/concrete/raindex/RaindexV6.sol";
+import {
+    CONTEXT_CALLING_CONTEXT_COLUMN,
+    CONTEXT_CALLING_CONTEXT_ROW_ORDER_HASH,
+    CONTEXT_CALLING_CONTEXT_ROW_ORDER_OWNER,
+    CONTEXT_CALLING_CONTEXT_ROW_ORDER_COUNTERPARTY,
+    CONTEXT_CALCULATIONS_COLUMN,
+    CONTEXT_CALCULATIONS_ROW_MAX_OUTPUT,
+    CONTEXT_CALCULATIONS_ROW_IO_RATIO,
+    CONTEXT_VAULT_INPUTS_COLUMN,
+    CONTEXT_VAULT_OUTPUTS_COLUMN,
+    CONTEXT_VAULT_IO_TOKEN,
+    CONTEXT_VAULT_IO_TOKEN_DECIMALS,
+    CONTEXT_VAULT_IO_VAULT_ID,
+    CONTEXT_VAULT_IO_BALANCE_BEFORE,
+    CONTEXT_VAULT_IO_BALANCE_DIFF
+} from "src/lib/LibRaindex.sol";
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed c4179e12b: This is additive, test-only mutation-hardening for the quote2 + calculateOrderIO logic of RaindexV6. It adds a single new test file under test/ and touches no source, audit, or generated artifact. Crucially it etches a fresh-compiled RaindexV6 subclass harness via type(Harness).runtimeCode (the same pattern as RaindexV6.vaultBalanceRevert.t.sol) so the calculate/quote source lines in src are actually exercised by the mutation pass rather than the bytecode-blind committed pointers used by the pre-existing quote coverage. The assertions are tight and discriminating: the calling-context calculating-context columns [orderHash, owner, counterparty] and per-vault [token, decimals, vaultId, balanceBefore, 0] rows, that each column reads its own token's TOFU decimals (input 6 vs output 12), the output-max vault cap as min(calculated, outputVaultBalance) (proven to fire above the balance, be inert below, floor at an empty vault, read the output balance row, and be min not max), the pre-fill ordering [capped max, ratio], the (IORatio, outputMax) read-back from stack slots 0/1, the CALCULATE_ORDER_MIN_OUTPUTS guard exactly at the < 2 boundary reverting UnsupportedCalculateOutputs(length), per-token TokenDecimalsReadFailure(token, outcome) for Inconsistent and ReadFailure on both input and output tokens with a Consistent+Initial positive control, and the quote2 dead-order early return (false,0,0) gated on liveness == ORDER_LIVE, the live (true, capped outputMax, IORatio) path, and the post-liveness TokenSelfTrade revert. Each behavior is pinned by a killed mutant from the campaign and exact values/events/selectors are asserted, so the suite discriminates correct from mutated behavior. No source changes. LGTM.

Issue #2543, opened 2026-04-10
Performance: fetch_order_trades query takes 1-2s per order hash on moderate databases


alastairong1:
## Problem

The `fetch_order_trades` SQL query in `crates/common/src/local_db/query/fetch_order_trades/query.sql` is very slow on databases with moderate data volume (~80MB, Base chain). A batch of 28 order hashes takes ~40 seconds, making trade lookups impractical for API use cases.

### 1. No connection pooling in `RusqliteExecutor`

`RusqliteExecutor::query_json` opens a fresh SQLite connection per query (`open_connection()` in `executor.rs:54-65`), including WAL mode setup, `busy_timeout` configuration, and custom function registration. When multiple queries run concurrently (e.g., fetching trades for multiple orders), each opens its own connection.

**Suggestion**: Use `r2d2-sqlite` or a similar pool, or cache a single connection per executor instance.

### 2. Missing composite indexes

The `take_trades` CTE joins `take_orders` to `order_add_events` on `(chain_id, orderbook_address, order_owner, order_nonce)`, but `idx_take_orders_owner` only covers `(chain_id, orderbook_address, order_owner)` — missing `order_nonce`.

Similarly, the final `vault_balance_changes` JOIN matches on `(owner, token, vault_id, block_number, log_index)` but the closest index (`idx_vbc_timestamp`) is ordered by `block_timestamp` rather than `block_number, log_index`.

**Suggestion**: Add covering indexes:
```sql
CREATE INDEX idx_take_orders_owner_nonce 
  ON take_orders (chain_id, orderbook_address, order_owner, order_nonce);

CREATE INDEX idx_vbc_block_log 
  ON vault_balance_changes (chain_id, orderbook_address, owner, token, vault_id, block_number, log_index);
```

### 3. N+1 query pattern

There's no batch variant of `fetch_order_trades` — callers must loop over order hashes one at a time. A single query with `WHERE order_hash IN (...)` would eliminate redundant CTE evaluations and connection overhead.

## Environment

- Database: ~80MB SQLite (Base chain, synced via `local-db-sync`)
- Concurrent queries: 28 order hashes via `join_all`
- Total wall time: ~40 seconds
- Per-query time: ~1.4 seconds average

## Workaround

We're currently bypassing the library's trade query and running optimized SQL directly against the SQLite file with connection pooling and batch lookups.

PR #2708, opened 2026-06-12
Reuse SQLite connections in RusqliteExecutor (perf, #2543)

thedavidmeister:
## What

Adds a simple connection-reuse pool to `RusqliteExecutor` in
`crates/common/src/local_db/executor.rs` so that the native local-db
executor stops opening a brand-new SQLite connection on every query.

Previously every `query_json` / `query_text` / `execute_batch` call ran
`open_connection` inside `spawn_blocking`: `Connection::open` plus
`journal_mode=wal`, `busy_timeout`, custom-function registration, and the
schema-version guard. A burst of concurrent order lookups (e.g.
`fetch_order_trades` across many orders) paid that full setup cost once
per query.

Now each method checks a ready connection out of a pool
(`Arc<Mutex<Vec<Connection>>>`, no new dependencies) — reusing an idle
one if present, otherwise opening a fresh one — and returns it to the
pool on success. A connection from a failed query/batch is dropped
instead of returned, so a connection left in a bad state never gets
reused. The mutex is held only for the brief pop/push, never across the
open or the query, so concurrency is preserved and the pool self-bounds
to peak concurrency (no explicit cap).

## Scope vs. issue #2543

Issue #2543 lists three root causes:

1. **No connection pooling** — fixed here.
2. **Missing indexes** — already addressed on `main`: `fetch_order_trades`
   reads from the materialized `derived_trades` table and
   `idx_derived_trades_order_hash_time ON derived_trades(chain_id,
   raindex_address, order_hash, block_timestamp DESC, block_number DESC,
   log_index DESC)` fully covers the query, so this PR deliberately does
   not touch indexes.
3. **No batch variant** — a batched `order_hash IN (...)` variant of
   `fetch_order_trades` remains a possible follow-up; out of scope here.

This is a Rust-only change in `crates/common`; the module is native-only
(`#[cfg(not(target_family = "wasm"))]`), so there are no wasm concerns.

## Tests

`cargo test -p raindex_common --lib local_db::executor` — all existing
executor tests pass, plus two new ones:

- `sequential_queries_reuse_a_single_pooled_connection` — runs several
  sequential `query_text` / `query_json` / `execute_batch` calls and
  asserts results are correct AND the pool never grows past one
  connection (fails if the connection is not returned to the pool).
- `failed_batch_does_not_return_connection_to_pool` — a failing batch
  drops its connection (pool ends empty) and the next successful query
  re-seeds the pool, proving clean recovery.

`cargo fmt` and `cargo clippy -p raindex_common --all-targets` are clean.

### DIFF
--- a/crates/common/src/local_db/executor.rs
+++ b/crates/common/src/local_db/executor.rs
@@ -941,4 +941,63 @@ mod tests {
             "a successful query after a failure re-seeds the pool"
         );
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    async fn concurrent_pool_access_does_not_deadlock() {
+        let temp_dir = TempDir::new().unwrap();
+        let db_path = temp_dir.path().join("concurrent-pool.db");
+
+        let exec = std::sync::Arc::new(RusqliteExecutor::new(&db_path));
+
+        // Seed a small table so the concurrent readers have real rows to return.
+        exec.query_text(&SqlStatement::new(
+            "CREATE TABLE nums (n INTEGER); INSERT INTO nums (n) VALUES (1), (2), (3), (4), (5);",
+        ))
+        .await
+        .unwrap();
+
+        #[derive(serde::Deserialize)]
+        struct SumRow {
+            total: i64,
+        }
+
+        // Fire a large number of concurrent operations through the real executor
+        // API. Each one races to check a connection out of the shared pool, run
+        // its query, and release it back, so the pool mutex is hammered under
+        // heavy contention. A regression that held the lock across the query (or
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed a1b4fd867: Reuse SQLite connections in RusqliteExecutor via a poison-recovered Arc<Mutex<Vec<Connection>>> pool (checkout/release with the lock held only for the O(1) pop/push, never across the query). Includes the concurrency stress test (100+ concurrent query_json/execute_batch under a 30s timeout) proving the single-mutex pool does not deadlock; the pool is deadlock-free by construction (one non-reentrant mutex, no lock across I/O). All required checks green. LGTM.

PR #2711, opened 2026-06-12
Add batched fetch_order_trades variant (IN clause)

### LINKED ISSUE #2543: Performance: fetch_order_trades query takes 1-2s per order hash on moderate databases

thedavidmeister:
## What

Adds a batched variant of `fetch_order_trades` so trades for many order hashes can be fetched in a single SQL query instead of N separate ones.

- `build_fetch_order_trades_batch_stmt(raindex_id, &[B256], start_ts, end_ts)` — emits the same query as the single-order builder but with `WHERE order_hash IN (...)`.
- `fetch_order_trades_batch(...)` wrapper in `raindex_client` — short-circuits an empty input to an empty result without touching the DB.
- `query.sql`: the order-hash predicate is now a `/*ORDER_HASH_CLAUSE*/` marker. The single-order builder fills it with `= ?3` via `bind_param_clause` (identical placeholder/param ordering to before, so the existing path is unchanged); the batch builder fills it with `IN ({list})` via the proven `bind_list_clause` pattern already used by `fetch_orders`.

## Why

Per #2543, there's no batch variant of `fetch_order_trades`, so callers must loop over order hashes one at a time. A single `WHERE order_hash IN (...)` query eliminates redundant CTE evaluations and per-query connection overhead (the N+1 pattern described in the issue). This is the batch half of the issue; it composes with connection pooling (#2708) as an orthogonal optimization. The existing `idx_derived_trades_order_hash_time` index covers the predicate.

No bytecode or deploy changes.

## Verification

- `cargo test -p raindex_common --lib local_db::query::fetch_order_trades` — 8 passed (includes new batch builder tests: IN-clause rendering + time filters, single-hash, empty-list clause removal).
- `cargo test -p raindex_common --lib local_db::query` — 195 passed (no regression in the single-order path or sibling query builders).
- `cargo check -p raindex_common --target wasm32-unknown-unknown --tests` — clean (new wasm wrapper tests compile).

Fixes #2543


thedavidmeister:
Reviewed 029b34956: explicit OrderHashFilter { All, These(&[B256]) } replacing implicit empty-means-all (All drops the clause; These([]) emits AND 1=0 = explicit none, never all); mutation-validated, 1077 tests green, rebased on green main, all checks green. LGTM.

PR #2714, opened 2026-06-13
ci: remove orphaned foundry.lock left over from soldeer migration

### LINKED ISSUE #2560: ci: verify foundry.lock revs match submodule pointers and exist upstream
## Problem

CI does not currently verify \`foundry.lock\` against either the submodule pointers in the tree or the upstream repos. It only clones submodules at whatever sha the tree records and runs the test suites against that. As a result a \`foundry.lock\` whose \`rev\` disagrees with the submodule pointer, or whose \`rev\` does not exist on the upstream repo at all, passes CI silently.

This shipped at #2558 — \`foundry.lock\` had \`lib/rain.interpreter.rev = 1ea41cd0ec4a8afd7dec9aeae0a65a55c95fac46\`, which does not exist on rainlanguage/rain.interpreter. The submodule pointer in the tree was the correct \`1ea41cd004a926e3a964ba5498654f3cd538f139\`, so all builds and tests passed. The breakage only surfaced when a downstream PR ran \`forge update lib/rain.interpreter\` and got \`fatal: unable to read tree\`. Fixed at #2559.

## Suggested check

Add a CI step that runs:

\`\`\`bash
nix develop -c forge update <each lib/*>
git diff --exit-code foundry.lock lib/
\`\`\`

For a tree whose lockfile is consistent with reachable upstream commits, \`forge update\` is a no-op — no diff, exit 0. If \`foundry.lock\`'s \`rev\` does not exist upstream, \`forge update\` fails with \`unable to read tree\` and CI fails. If \`foundry.lock\`'s \`rev\` is a real but stale commit (not matching the tree pointer), \`forge update\` rewrites the submodule pointer to match, \`git diff\` is non-empty, and CI fails.

The check should iterate the four submodules in \`foundry.lock\`: \`lib/rain.deploy\`, \`lib/rain.interpreter\`, \`lib/rain.raindex.interface\`, \`lib/sushixswap-v2\`.

## Out of scope

Whether the upstream commit is on the dep's main branch (i.e., catching pins to closed/abandoned branches) is a different concern — the proposed check only verifies reachability and consistency, not branch policy.

thedavidmeister:
## What

Removes the orphaned `foundry.lock` (the old git-submodule dependency lockfile) and its two now-dangling housekeeping references in `REUSE.toml` and `.soldeerignore`.

## Why

Issue #2560 asked for a CI step that runs `forge update` on the four git submodules recorded in `foundry.lock` (`lib/rain.deploy`, `lib/rain.interpreter`, `lib/rain.raindex.interface`, `lib/sushixswap-v2`) and asserts `git diff --exit-code`, to keep the lockfile revs consistent with the submodule pointers and reachable upstream. That was the correct fix for the bug shipped at #2558 — a `foundry.lock` rev that did not exist upstream passed CI silently and only broke when a downstream `forge update` hit `fatal: unable to read tree`.

Since the issue was filed, **#2586** ("migrate raindex to soldeer for dependency management", merged 2026-05-29) deleted all four git submodules and `.gitmodules`. Every Solidity dependency is now Soldeer-managed via `soldeer.lock` + `[dependencies]` in `foundry.toml` (`foundry.toml` even states: *"All Solidity dependencies are Soldeer-managed; no git submodules remain."*).

So the submodules the proposed `forge update` check would iterate **no longer exist** — the literal check is moot. But `foundry.lock` itself was *not* removed by #2586 and is now a dead, unreferenced artifact pointing at submodules that are gone (sushixswap-v2 was dropped entirely; the other three became the soldeer deps `rain-deploy` / `rain-interpreter-interface`+`rainlang` / `raindex-interface`).

The smallest correct fix that resolves the exact inconsistency the issue worries about — a dependency lockfile that disagrees with the tree and can silently pass CI — is to **delete the orphaned lockfile** so there is no longer any such file to drift. No CI job reads it: the Foundry cache key in rainix's `rainix-sol-{test,static}.yaml` hashes `soldeer.lock` (not `foundry.lock`), and dependencies come from `forge soldeer install`.

Also drops the two now-dangling housekeeping references to the file:
- `REUSE.toml` annotation entry (it kept `reuse lint` happy for the file).
- `.soldeerignore` exclude line (it excluded the file from the soldeer publish zip).

## Verification

Run in `nix develop github:rainlanguage/rainix#sol-shell`:

- `reuse lint` → compliant, 1139/1139 files, 0 errors, 0 unused licenses (confirms the `REUSE.toml` edit is clean).
- `forge soldeer install` → dependencies installed from `soldeer.lock`; `foundry.lock` is not consulted.
- `forge build` → compiles clean (exit 0; only pre-existing `unsafe-typecast` lint *warnings* in test files, unrelated to this change).

Fixes #2560


thedavidmeister:
Reviewed ee63d4675: Removes the orphaned foundry.lock left over from the submodule->soldeer migration (Solidity deps are pinned via foundry.toml [dependencies] + soldeer.lock now), and drops its stale REUSE.toml / .soldeerignore entries. CI/housekeeping only - no src, bytecode, or subgraph impact. All checks green. LGTM.

PR #2716, opened 2026-06-13
Add unified RaindexVault.getCalldatas() method

### LINKED ISSUE #1976: Combine all of the calldata fns we use for the vaults into a single function in RaindexClient
Currently we have multiple calldata generation fns on the vault instance. Instead of having separate, we could unify those in a single function called `vault.getCalldatas()`. 

@hardyjosh made a comment on this https://github.com/rainlanguage/rain.orderbook/pull/1956#discussion_r2191022749

thedavidmeister:
## What

Adds a unified `RaindexVault.getCalldatas(amount)` method that consolidates the three separate vault calldata functions (`getDepositCalldata`, `getWithdrawCalldata`, `getApprovalCalldata`) into a single call.

It returns a new `RaindexVaultCalldatas` struct:

```ts
interface RaindexVaultCalldatas {
  approval?: Hex; // undefined when no approval is needed
  deposit: Hex;
  withdraw: Hex;
}
```

The on-chain allowance is read only once. `approval` is `undefined`/`None` when the raindex contract already has a sufficient allowance for the requested amount, so callers no longer have to treat "no approval needed" as an error (the pattern that `handleVaultDeposit` currently works around). The approval-building logic is factored into a shared `build_approval_calldata` helper so `get_approval_calldata` and `get_calldatas` stay in sync.

## Why

The three calldata functions are always used together (e.g. the webapp deposit flow calls `getApprovalCalldata` then `getDepositCalldata`). Combining them into one method is the API improvement requested in the issue (and in the referenced `hardyjosh` review comment, which suggested returning `Option` rather than an error for the no-approval case).

This is purely additive: the existing individual functions are unchanged, so there are no bytecode or subgraph changes.

## Verification

- `cargo test -p raindex_common --lib raindex_client::vaults` — all 28 tests pass, including the new `test_get_vault_calldatas`.
- `cargo clippy -p raindex_common --lib` — clean.
- Rebuilt the wasm bindings (`npm run build -w @rainlanguage/raindex`); the generated `RaindexVaultCalldatas` interface and `getCalldatas` method appear in the `.d.ts` as expected, and `tsc` check passes.
- `vitest run test/js_api/raindexClient.test.ts` — all 35 tests pass, including three new `getCalldatas` cases (approval needed, approval not needed, zero amount).
- `prettier --check` clean on the changed test file.

Fixes #1976

### DIFF
--- a/crates/common/src/raindex_client/mod.rs
+++ b/crates/common/src/raindex_client/mod.rs
@@ -453,8 +453,6 @@ pub enum RaindexError {
     ZeroAmount,
     #[error("Negative amount")]
     NegativeAmount,
-    #[error("Existing allowance")]
-    ExistingAllowance,
     #[error(transparent)]
     WritableTransactionExecuteError(#[from] WritableTransactionExecuteError),
     #[error(transparent)]
@@ -628,9 +626,6 @@ impl RaindexError {
             RaindexError::NegativeAmount => "Amount cannot be negative".to_string(),
             RaindexError::WritableTransactionExecuteError(err) => {
                 format!("Failed to execute transaction: {}", err)
-            }
-            RaindexError::ExistingAllowance => {
-                "There is already an allowance for this vault".to_string()
             }
             RaindexError::DepositArgsError(err) => {
                 format!("Failed to create deposit arguments: {}", err)
--- a/crates/common/src/raindex_client/vaults.rs
+++ b/crates/common/src/raindex_client/vaults.rs
@@ -392,74 +392,6 @@ impl RaindexVault {
         Ok(())
     }
 
-    /// Generates transaction calldata for depositing tokens into a vault
-    ///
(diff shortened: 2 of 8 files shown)


thedavidmeister:
Reviewed ba76bc98c: Add unified RaindexVault.getCalldatas() method — consolidates the per-vault calldata methods into vault.getCalldatas() per #1976 (removed getDeposit/Withdraw/ApprovalCalldata, migrated the webapp withdraw flow, dropped the orphan ExistingAllowance variant). All rust/wasm/vitest/js-binding checks GREEN. The only reds are pre-existing main sol issues this rust/wasm change cannot affect (rainix-sol/static = multi-contract files fixed by #2731; rainix-sol/test = the 0_1_7 missing-deploy-constants test + the OrderNoSources quote tests, both fixed separately). LGTM.

PR #2718, opened 2026-06-13
Restore Auction and Dynamic spread full deployment tests (#2037)

### LINKED ISSUE #2037: Bring back commented out tests in fullDeployment.test.ts once all strats are supported in rain.strategies

thedavidmeister:
## What

Re-enables the two commented-out UI deployment tests in
`packages/webapp/src/routes/deploy/[orderName]/[deploymentKey]/fullDeployment.test.ts`:

- **Auction order** (`auction-dca`)
- **Dynamic spread order** (`dynamic-spread`)

## Why

These cases were commented out because they depended on undefined
`auctionOrder` / `dynamicSpreadOrder` dotrain constants and the removed
`RaindexOrderBuilder.newWithDeployment` API. Both strategies are now served
from the rain.strategies registry, so the tests are rewritten to follow the
established registry-based pattern already used by the "Fixed limit order"
test — looking up deployment/order details via `registry.getDeploymentDetails`
and `registry.getAllOrderDetails`, driving the page through the real registry,
and rebuilding the expected calldata with `registry.getOrderBuilder`.

Field/binding names, select-token keys, and vault id inputs were verified
against the live strategy definitions:

- **auction-dca**: `output` / `input` select tokens; required fields
  `time-per-amount-epoch`, `amount-per-epoch`, `max-trade-amount`,
  `min-trade-amount`, `baseline`, `initial-io`. One output + one input → two
  vault id inputs.
- **dynamic-spread**: `token1` / `token2` rotated tokens, each listed as both
  an input and an output, so there are **four** vault id inputs (the old
  commented draft only handled two); `amount-is-fast-exit` /
  `not-amount-is-fast-exit` presets plus `initial-io`, `max-amount`,
  `min-amount`.

## Verification

Run in `nix develop .#wasm-shell` after building `@rainlanguage/raindex` and
`@rainlanguage/ui-components`:

- `npm run test:unit -w @rainlanguage/webapp -- --run fullDeployment.test.ts`
  → **9 passed (9)**, including `Auction order` and `Dynamic spread order`
  (confirmed twice).
- `eslint` on the file → clean.
- `svelte-check` → 0 errors / 0 warnings (with `.env` populated; the only
  diagnostics without it are the pre-existing `PUBLIC_WALLETCONNECT_PROJECT_ID`
  env errors, unrelated to this change).

Fixes #2037

### DIFF
--- a/packages/webapp/src/routes/deploy/[orderName]/[deploymentKey]/fullDeployment.test.ts
+++ b/packages/webapp/src/routes/deploy/[orderName]/[deploymentKey]/fullDeployment.test.ts
@@ -397,313 +397,368 @@ describe('Full Deployment Tests', () => {
 		{ timeout: 60000, retry: DEFAULT_MAX_RETRIES }
 	);
 
-	// TODO: Issue #2037
-	// it(
-	// 	'Auction order',
-	// 	async () => {
-	// 		mockPageStore.mockSetSubscribeValue({
-	// 			data: {
-	// 				dotrain: auctionOrder,
-	// 				deployment: {
-	// 					key: 'base'
-	// 				},
-	// 				orderDetail: {
-	// 					name: 'Auction'
-	// 				}
-	// 			}
-	// 		});
+	it(
+		'Auction order',
+		async () => {
+			const auctionDeploymentDetails = registry.getDeploymentDetails('auction-dca');
+			if (auctionDeploymentDetails.error) {
+				throw new Error('Failed to get deployment details');
+			}
+			const deployment = auctionDeploymentDetails.value.get('base') as NameAndDescriptionCfg;
+			const auctionOrderDetail = registry
+				.getAllOrderDetails()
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed c8fdc4502: Restores the two parked fullDeployment.test.ts UI deployment tests (auction-dca, dynamic-spread) that were commented behind a TODO marker because they depended on the removed inline auctionOrder/dynamicSpreadOrder dotrain constants and the removed RaindexOrderBuilder.newWithDeployment API. Both strategies are now served from rain.strategies, and the rewrite follows the established registry-based pattern (getDeploymentDetails / getAllOrderDetails / getOrderBuilder) already used by the Fixed-limit test. Correctly models dynamic-spread's four vault-id inputs (token1/token2 each listed as both input and output), which the old two-input draft got wrong. Test-only change; fullDeployment.test.ts is 9/9 green and every required check plus rainix-sol test/static/legal and copy-artifacts are green. LGTM.

PR #2719, opened 2026-06-13
Strip RPC URLs from generated dotrain to prevent API token leak (#2165)

### LINKED ISSUE #2165: Replace rpc urls for the order before publishing dotrain
If a user has some api tokens in their rpc definitions, creating an order would leak this to everyone. We should be able to remove the rpc urls and put local/webapp settings in place of those

thedavidmeister:
Fixes #2165

## Problem

`generate_dotrain_for_deployment()` (`crates/common/src/dotrain_order.rs`) copied the deployment network's `networks` section verbatim into the dotrain it produces. That generated dotrain is shared with others and embedded into on-chain order metadata (via `AddOrderArgs::new_from_deployment`), so any private API token baked into a user's RPC URL — whether in a path segment (`/v2/<key>`) or a query string (`?apikey=<token>`) — leaked to everyone who could read the order.

## Fix

Following the existing vault-id stripping pattern in the same function, a new `strip_rpcs_from_network` helper replaces every RPC URL in the emitted network entry with a non-secret placeholder (`https://rpc.example.com`) before the dotrain is returned.

The placeholder is a parseable URL, so the required, non-empty `rpcs` array stays structurally valid and the generated dotrain still parses end-to-end. Consumers (the webapp / local settings) inject their own RPCs at use time, which is exactly what the issue asks for ("remove/strip the rpc urls and put local/webapp settings in place of those if needed").

## Test

Adds `test_generate_dotrain_for_deployment_strips_rpc_api_tokens`, a discriminating test that:
- embeds two distinct secrets in the network's RPC URLs (a path-style API key and a query-string token);
- asserts neither secret string, nor either full secret-bearing URL, appears anywhere in the generated dotrain (the core security assertion);
- asserts each of the two rpc entries is replaced by the placeholder;
- asserts the generated dotrain still parses via `DotrainOrder::create`.

Verified the test FAILS against the unpatched function (panics with "path API key leaked into generated dotrain") and PASSES with the fix.

## Verification

- `cargo test -p raindex_common --lib dotrain_order::` — 18 passed, 0 failed.
- `cargo test -p raindex_common --lib raindex_order_builder::` — 92 passed, 0 failed (these consume `generate_dotrain_for_deployment` and recompute their expected hashes, so they self-consistently track the content change).
- `cargo fmt` applied.


thedavidmeister:
🤖 ai:producer
Producer note: all three red checks are pre-existing MAIN red, not PR-caused — rainix no-skips static gate (vm.skip x5 in test/lib/deploy, fails rainix-sol/static + rs-static) and missing 0.1.13 soldeer deploy-constant pins (testAllPublishedSoldeerTagsHaveAFullConstantSuite; 687 other sol tests pass). Both are fixed by #2804 (green, ai:ready); this PR greens on a branch update once it lands. flag-blocked-on was refused by the guard (human decision present), so recording the classification here. Parked under human:reject — no rework note to execute.

thedavidmeister:
🤖 ai:producer

State change: the dependency this PR was recorded as waiting on is dead. My earlier note said the three red checks (rainix no-skips static gate; missing 0.1.13 soldeer deploy-constant pins) were pre-existing MAIN reds fixed by rainlanguage/raindex#2804, and that this PR would green on a branch update once that landed.

rainlanguage/raindex#2804 was **closed without merging**, so that fix never landed and those main reds still stand. This PR is parked under `ai:reject` with no rework note, so I am not re-attempting it — recording the correction only, so the earlier note is not read as a live path to green.

Issue #2327, opened 2025-11-19
add retry when deposit tx timeouts


rouzwelt:
![Image](https://github.com/user-attachments/assets/916db765-8cb3-475f-a7d9-4a54274ea02d)

timed out promises will cause such errors in the gut, probably we can setup a retry logic so that such errors are resolved and occurs less often

---

## Updated scope (2026-06-18): detect the timeout via a TYPED error discriminant, not the message text

The retry mechanic itself is simple — retry the SDK indexing call a few times on a transient timeout before surfacing a hard failure. What this issue now requires is that **a "timeout" be detected from a typed/structured error signal, not from the human-readable message text.**

A first attempt — PR #2720 (TS-only) — gated the retry on a substring match of the message:

```ts
result.error?.readableMsg?.toLowerCase().includes("timeout")
```

That approach was rejected as fragile:

- **Silent false-negative (the dangerous one):** if the SDK ever rewords the message ("timed out", "deadline exceeded", "request expired", "no subgraph response", …), detection returns `false`, the retry never fires, and this fix silently regresses — with no test catching it, because the tests hardcode the `"...timeout..."` message string.
- **False-positive → over-retry:** any unrelated error whose message merely contains "timeout" (e.g. an RPC-timeout config error, a revert reason naming a timeout param) gets retried 3× before the user sees the real failure.
- It couples runtime control-flow to presentation copy and rests on an undocumented Rust↔TS string convention.

### Requirement

1. The SDK (`@rainlanguage/raindex`, i.e. the Rust/wasm bindings) exposes a **typed, structured discriminant** on a `WasmEncodedResult` error — an error `kind`/`code` enum with an explicit timeout variant (e.g. `SubgraphTimeout`) — rather than only a human-readable `readableMsg`.
2. The ui-components retry gate (`isTimeoutResult` in `TransactionManager.ts`) **and** the existing `SUBGRAPH_TIMEOUT_ERROR` labeling both branch on that typed field — never on `readableMsg` text.
3. Tests assert against the typed discriminant (ideally pinning that the SDK's real timeout path yields that variant), so a message rewording cannot silently disable the retry.

This is a cross-language change (Rust/wasm error encoding + TS consumption), larger than the TS-only #2720. The retry/backoff mechanics from #2720 can be reused as-is — only the **detection** must move from prose-matching to a typed signal.

thedavidmeister:
[...]

PR #2720, opened 2026-06-13
Retry SDK indexing on transient timeout before failing

thedavidmeister:
Depends on #2327.

## Problem

Timed-out indexing promises for deposit / withdraw / order transactions surfaced as a hard failure on the very first timeout, even though the timeout is transient (the subgraph just hasn't indexed the tx yet). This produced spurious "failed" toasts for transactions that actually succeeded on-chain.

## Change (TS / ui-components only)

`createSdkIndexingFn` in `packages/ui-components/src/lib/providers/transactions/TransactionManager.ts` now retries the SDK indexing call up to `DEFAULT_MAX_RETRIES` (3) total attempts on a timeout result, using the existing `retry()` backoff, before marking the transaction failed with `SUBGRAPH_TIMEOUT_ERROR`. This covers every flow routed through `createSdkIndexingFn`: deposit, withdraw, vaults-withdraw-all, add-order, remove-order, and take-order.

Only timeouts are retried. A successful result, a non-timeout hard error result, and a rejection thrown by the SDK call itself are all surfaced after a single attempt (unchanged fail-fast behavior).

A `retry()` util (mirroring the one in `packages/webapp/src/lib/retry.ts`) is added to ui-components because `TransactionManager` lives in ui-components and cannot import from webapp (webapp depends on ui-components, not the reverse).

## Tests (all discriminating — verified to fail against the pre-change code)

New `createSdkIndexingFn > timeout retry behavior` block plus updated existing timeout tests:

- retries exactly 3 times on a persistent timeout, then `SUBGRAPH_TIMEOUT_ERROR` + `onError` (catches no-retry: 1 call).
- recovers and succeeds (`SUCCESS` + `onSuccess`, no error state) when a retry after a timeout returns valid data (catches no-retry: fails on first timeout).
- does NOT retry a non-timeout hard error (exactly 1 call) — guards against over-retrying.
- does NOT retry a thrown SDK rejection (exactly 1 call) — guards against over-retrying.
- The two integration-level `should handle SDK timeout error in awaitIndexingFn` tests (remove-order, deposit) were updated to a persistent-timeout mock + `toHaveBeenCalledTimes(3)`, reflecting the new retry behavior.

Reverting only the source change makes exactly these 4 retry assertions fail, confirming they discriminate.

## Verification

- `TransactionManager.test.ts`: 55/55 pass (39ms under fake timers, no real backoff waits).
- Full `@rainlanguage/ui-components` suite: 634 tests pass / 100 test files pass. The 2 failing test files (`CodeMirrorDotrain.test.ts`, `License.test.ts`) are pre-existing, environment-caused (`Failed to resolve entry for package "@rainlanguage/ui-components"` — the package self-entry is built by CI's `npm run build -w @rainlanguage/ui-components`); confirmed identical on clean `main` with my changes stashed.
- Lint (`prettier --list-different` + `eslint`): clean. Prettier format: no changes.


thedavidmeister:
To be clear on disposition: this PR is **not** being discarded — the retry/backoff mechanic and its tests are sound. It's **blocked by #2327**, which has been re-scoped to add a *typed* timeout discriminant (`error.kind`/`code`) to the SDK.

The only part of this PR that changes once #2327 lands is the detection: `isTimeoutResult` (and the `SUBGRAPH_TIMEOUT_ERROR` labeling) branch on the typed field instead of substring-matching `readableMsg`. Until that swap the current substring detection shouldn't merge as-is (that's the standing review reject) — but the PR stays **open** as the retry implementation waiting on its dependency. The reject clears when #2327 lands and this PR adopts the typed signal.

thedavidmeister:
🤖 ai:vetter
Reviewed 2791a97fb91d2e35210a592f003adfe9449dd8da: ready — TimeoutRetrySignal correctly separates transient timeouts from hard errors; thrown SDK rejections ca
cost 200 — TS retry logic with sentinel error class

thedavidmeister:
Human design hold (adversarial sweep, David-authorized 2026-07-09). Parks a design question. `isTimeoutResult()` classifies timeouts by `result.error?.readableMsg?.toLowerCase().includes("timeout")` — substring-matching the human-readable error message, the exact anti-pattern the org ruling forbids (classify by a typed `kind`/`code`, never `readableMsg`) and that issue #2327 (updated 2026-06-18) names this PR and rejects. The `TimeoutRetrySignal` class in the title is post-classification retry plumbing, not a typed discriminant. Design question: detect timeout via a typed SDK/wasm discriminant (a `SubgraphTimeout` variant, per #2327) vs the current message-substring — the latter silently regresses if the SDK rewords "timeout" and over-retries any error whose text contains it. Not merged pending the ruling.

thedavidmeister:
🤖 ai:producer
Producer note: the three red checks are MAIN-STATE, not PR-caused — missing 0.1.13 soldeer deploy-constant pins (rainix-sol/test) and the no-vm.skip gate on 5 pre-existing main files (rainix-sol/static + rs-static); identical failures on main HEAD. Separate from the open human design hold (typed discriminant, depends on #2327), which remains the real gate.

thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens no source read — the ai:design label's provenance decides this routing, not the diff
Reviewed 32ef31c7b456b75f87a9a9ce912c8079ad109440: needs-work — ROUTING FIX, not a code defect (design doctor, https://github.com/rainlanguage/issue-pr-cron/issues/241): this PR carried ai:design with NO trusted comment raising a design question at its head, which is a state no actor consumes — the human's queue withholds it (there is no claim to present) and every AI actor skips it. It is routed here so it has an owner again. Two moves, both yours: if the PR still needs work, push it — the question was superseded or never recorded, and the push IS the transition. If a genuine design question remains, re-raise it with `pr-review-report flag-design <owner/repo> <pr> "<the question>"`, which puts the row back in the human's queue with a claim behind it. Do not push a no-op commit to clear this.

PR #2722, opened 2026-06-13
Add "remove and withdraw all" option to order removal (#2024)

### LINKED ISSUE #2024: Allow withdrawing at the same time as canceling an order
withdraws on order removal shouldn't be automatic as orders can share vaults but order removal can offer the option to also withdraw tokens from the vault. the remove button could be replaced with a context menu button with "remove" and "remove and withdraw all vaults"

thedavidmeister:
Fixes #2024

## What the issue asked

Allow withdrawing an order's vault balances at the same time as removing the
order. Per the issue, withdrawal must **not** be automatic (orders can share
vaults), so the Remove button should become a context menu offering "remove"
and "remove and withdraw all vaults".

## What changed (TS/webapp only — no contract/bytecode changes)

- **`OrderDetail.svelte`** — when the new optional `onRemoveAndWithdrawAll`
  callback is provided, the standalone Remove button is replaced by a dropdown
  with two items: **Remove** (unchanged behaviour) and **Remove and withdraw all
  vaults**. When the callback is not supplied the original single Remove button
  renders, so existing callers are unaffected.
- **`handleRemoveOrderAndWithdrawAll.ts`** (new webapp service) — mirrors
  `handleVaultsWithdrawAll`, but generates a single
  `multicall([removeOrder, withdraw])` so the order removal and the vault
  withdrawals execute atomically in one transaction. The withdraw calldata comes
  from `vaultsList.getWithdrawCalldata()` and is only appended when the order
  actually has withdrawable (positive-balance) vaults — an order whose vaults
  are all empty is still removed. The multicall is built with viem's
  `encodeFunctionData` against the Raindex `multicall(bytes[])` ABI (selector
  `0xac9650d8`, inherited from OpenZeppelin `Multicall`).
- **order detail `+page.svelte`** — wires the new `onRemoveAndWithdrawAll`
  prop to the new service.

## Tests (all discriminating)

- **`handleRemoveOrderAndWithdrawAll.test.ts`** (7 tests):
  - asserts the **exact** multicall calldata bytes for `[removeOrder, withdraw]`
    and decodes it to prove the flat order (a swapped order or a missing withdraw
    changes the bytes — verified by mutation);
  - asserts an order with no withdrawable vaults produces a **remove-only**
    multicall and never even calls `getWithdrawCalldata`;
  - asserts the combined modal title, the remove-order transaction created on
    confirm with the right args, and that remove-calldata / withdraw-calldata /
    withdrawable-lookup errors each toast and abort without opening the modal.
- **`OrderDetail.test.ts`** (4 new tests): the dropdown replaces the bare button
  when `onRemoveAndWithdrawAll` is set; the Remove item calls `onRemove` (and not
  the other); the Remove-and-withdraw item calls `onRemoveAndWithdrawAll` with
  the order and its `vaultsList` (and not `onRemove`) — verified by mutation; and
  the dropdown is hidden for inactive orders.

## Verification

- ui-components full suite: **640 passed** (incl. 31 OrderDetail tests).
- webapp full suite: **187 passed** (incl. 7 new service tests).
- `npm run check`: ui-components 0 errors; webapp 2 pre-existing errors only
  (`PUBLIC_WALLETCONNECT_PROJECT_ID` missing env var, in untouched files).
- Lint clean on all touched files; Prettier applied.

## Summary by CodeRabbit

* **New Features**
  * Added a combined “Remove and withdraw all vaults” option for eligible orders.
  * When available, the order details view now shows a dropdown with separate removal actions.

* **Bug Fixes**
  * Improved handling for inactive orders so removal actions are not shown.
  * Streamlined transaction flow so removal and vault withdrawal can be submitted together when applicable.

* **Tests**
  * Added coverage for the new removal dropdown, combined transaction flow, and error handling.


thedavidmeister:
screenshot

UI screenshot: "remove and withdraw all" option on order removal

thedavidmeister:
🤖 ai:vetter
Reviewed 4d47d8737fd430db22a73e5692de8a0d267d2793: ready — multicall([removeCalldata,withdrawCalldata]) uses correct OZ selector 0xac9650d8; empty-vault case c
cost 300 — multicall encoding plus exact calldata byte tests

thedavidmeister:
🤖 ai:vetter
Reviewed 114ad0a6e01557b99fac2b0b75823d23e17d0ff5: ready — closes #2024 — dropdown with remove / remove-and-withdraw-all exactly as issue asks (non-automatic); TS-only multicall([removeOrder, withdraw]) batching is sound (remove does not touch vaults, withdraw is target-min on-chain, OZ Multicall non-payable); fail-closed error paths; exact-calldata-bytes tests + component wiring tests both discriminating; reds are pre-existing main reds
cost 355 — webapp multicall feature, exact-bytes tests

thedavidmeister:
🤖 ai:producer
Producer note: all three red checks are MAIN-STATE, not PR-caused — rainix-sol/test fails testAllPublishedSoldeerTagsHaveAFullConstantSuite (soldeer tag 0.1.13 published without its deploy-constant pins, byte-identical on main), and rainix-sol/static + rs-static fail the no-vm.skip gate on 5 pre-existing test/lib/deploy files on main (e.g. LibRaindexDeployTaggedConstants.t.sol:24). Nothing fixable from this branch; greens when main pins 0.1.13 and resolves the vm.skip gate.

thedavidmeister:
🤖 ai:producer
Blocked-on: Reds are the main-state trio (rainix-sol static/test + rs-static: banned vm.skip usages + missing 0.1.13 deploy constants) inherited by every open PR — same failing-check signature as the #2798 sample verified against the logs. Fix is #2804; greens after it lands and this branch updates.

PR #2726, opened 2026-06-13
Gate GH_COMMIT_SHA env! read behind #[cfg(test)]

### LINKED ISSUE #650: elide testing env vars from prod builds
## Motivation

Currently we have to set envs unconditionally due to env! but really they are only needed for testing

## Solution

Elide testing envs with `#[cfg(test)]` so they are only needing to be set during testing.

thedavidmeister:
Fixes #650.

## Problem

`crates/common/src/lib.rs` read `COMMIT_SHA` at compile time via an
unconditional `env!("COMMIT_SHA", "$COMMIT_SHA not set.")`. That made the env var
a hard compile-time input for **every** build — production included — even though
`GH_COMMIT_SHA` has no production consumer anywhere in the workspace (no Rust, JS,
or wasm code reads it). The constant only exists for test-side traceability.

## Change

- `crates/common/src/lib.rs`: gate `pub const GH_COMMIT_SHA` (and its `env!`
  read) behind `#[cfg(test)]`. Production builds now neither read the env var nor
  carry the constant; only test builds require `COMMIT_SHA` to resolve at compile
  time.
- `crates/common/build.rs`: keep deriving `COMMIT_SHA` (CI's `github.sha`, else
  `git rev-parse HEAD`, else `"unknown"`) so test builds still get a value — a
  build script can't tell the test profile from a prod build, so the derivation
  stays unconditional and is harmless for prod, which no longer reads the var.
  Comment updated to describe this.
- `crates/common/ARCHITECTURE.md`: updated the three `GH_COMMIT_SHA` references to
  describe the now test-only binding.

## Test

Added `commit_sha_tests::gh_commit_sha_matches_commit_sha_env` in `lib.rs`,
asserting:
- `GH_COMMIT_SHA == env!("COMMIT_SHA")` — the constant carries the exact value
  `build.rs` resolved.
- `GH_COMMIT_SHA != "$COMMIT_SHA not set."` — the unset-sentinel never leaks.
- `!GH_COMMIT_SHA.is_empty()`.

Discrimination verified by mutation: replacing the constant with a hardcoded
`"MUTANT_WRONG_SHA"` makes the test fail (`left: "MUTANT_WRONG_SHA"` vs
`right: <git HEAD sha>`); restoring it passes.

## Verification

- Prod build: `cargo build -p raindex_common` (no `#[cfg(test)]`) — compiles
  clean with the constant elided, proving nothing else references it.
- Test build: `cargo test -p raindex_common --lib commit_sha` — `1 passed`
  (1073 other lib tests filtered out, suite intact).
- `cargo fmt --all` clean.


thedavidmeister:
Reviewed d763e131e: approved by maintainer; rebased onto post-audit main (5c0c914c6), all required + sol checks green, mergeable, CodeRabbit clean. Merging.

PR #2727, opened 2026-06-13
fix(lsp): don't flag elided bindings set in front matter (#587)

### LINKED ISSUE #587: Codemirror shouldn't error if an elided binding is actually set in the front matter
telegram-cloud-photo-size-4-5818672241100046206-y

this shouldn't error if the binding is set in the frontmatter

thedavidmeister:
Fixes #587

## Problem

The CodeMirror rainlang linter reported an "elided binding" error even when that binding *is* actually set in the front matter. `DotrainAddOrderLsp::problems` builds a `RainDocument` from only the rainlang body after the front matter splitter, so dotrain never sees the scenario bindings and flags `#name !...` elided bindings that the front matter does in fact provide. These errors are noisy false positives in the editor.

## Fix

`DotrainAddOrderLsp::problems` now collects the union of every binding key set under any (including nested) scenario in the front matter YAML, and filters out `ErrorCode::ElidedBinding` problems whose binding name is in that set — applied to both the bindings-problems path and the `ComposeError::Problems` path.

- Elided bindings that are genuinely unset are still reported (e.g. `fixed-io`).
- If the front matter fails to parse, the collected set is empty, so nothing is suppressed and behavior is unchanged.
- No deployed-contract bytecode is touched.

## Tests

- `test_problems` (updated): `fixed-io-output-token` is elided **and** set in the `flare` scenario bindings, so its error is suppressed; `fixed-io` is elided and unset, so its error remains. Asserts the exact full problem-message vector and that no message mentions `fixed-io-output-token`.
- `test_elided_binding_set_in_frontmatter_is_not_reported` (new): a `LocalEvm` round trip where an elided body binding `bound` is provided by the `flare` scenario; asserts the composed result has zero problems.
- `test_frontmatter_scenario_binding_keys_unions_nested_scenarios` / `..._ignores_non_binding_keys` (new): the scenario binding-key collector picks up nested-scenario bindings and never treats scenario names / reserved keys as bindings.
- `test_elided_binding_name_extraction` (new): the binding-name parser handles hyphenated names and descriptions that themselves contain quotes.

## Verification

- `cargo test -p raindex_common --lib`: **1077 passed, 0 failed, 0 ignored, 0 filtered** (whole common-crate lib suite green on this branch; LocalEvm tests spin up a local anvil, no external RPC needed).
- `cargo fmt` (only this file reformatted) and `cargo clippy -p raindex_common --all-targets` clean.


thedavidmeister:
human-approved; merging

Issue #1840, opened 2025-05-16
MockComponent is throwing invalid array length errors


hardingjam:
Causes the tests to hang here:

![Image](https://github.com/user-attachments/assets/f5e26a01-ddb8-4fdb-9d32-bd45f1f1aa80)

For 30s+ before this error appears:

```
RangeError: Invalid array length
 ❯ make_dirty ../../node_modules/svelte/src/runtime/internal/Component.js:75:42
     76| 
     77| /** @returns {void} */
     78| function make_dirty(component, i) {
       |      ^
     79|  if (component.$$.dirty[0] === -1) {
     80|   dirty_components.push(component);
 ❯ ../../node_modules/svelte/src/runtime/internal/Component.js:134:17
 ❯ Array.<anonymous> src/lib/__mocks__/MockComponent.svelte:74:92
 ❯ Object.set ../../node_modules/svelte/src/runtime/store/index.js:51:29
 ❯ Module.set_store_value ../../node_modules/svelte/src/runtime/internal/utils.js:301:8
 ❯ Object.$$self.$$.update src/lib/__mocks__/MockComponent.svelte:5:2
 ❯ update ../../node_modules/svelte/src/runtime/internal/scheduler.js:125:6
 ❯ Module.flush ../../node_modules/svelte/src/runtime/internal/scheduler.js:88:5
 ❯ Module.init ../../node_modules/svelte/src/runtime/internal/Component.js:159:25
```

When using multiple instances of MockComponent in a test

PR #2728, opened 2026-06-13
Pass distinct props (testid) to each MockComponent instance

### LINKED ISSUE #1842: Find a way to pass different props to each instance of MockComponent in test files
See #1840

The previous reactive props system was causing this error, but it would be good to be able to pass testid prop to each instance of MockComponent, in test files that use more than one of those.

thedavidmeister:
Fixes #1842

## Motivation

The `ui-components` mock at `packages/ui-components/src/lib/__mocks__/MockComponent.svelte` still carried the reactive props store (`$: $props = $$props`) that #1840 identified as the cause of the `RangeError: Invalid array length` crash/hang when a test mounts more than one instance. PR #1841 only fixed the `webapp` copy of the mock; this one was left untouched. There was also no way to give each instance distinct props (e.g. a `testid`) so multiple instances could be told apart.

## Solution

Rework the `ui-components` mock so it:

- Mirrors the received props into the shared `props` store from `afterUpdate` instead of a reactive `$:` block, so it never re-enters the component update cycle that exhausted the dirty array across multiple instances (the #1840 crash/hang).
- Exports a `testid` prop (default `"mock-component"`) that drives `data-testid`, so a test mounting more than one instance can give each a distinct id and query them individually. Remaining props are still spread onto the element and into the store, so existing attribute- and store-based assertions keep working (`License`, `CodeMirror*`, `TanstackLightweightChartLine`).

## Tests

Added `MockComponent.test.ts` (plus a `MockComponent.test.svelte` harness) with discriminating assertions:

- default `data-testid` is `mock-component`;
- a `testid` prop overrides `data-testid` and removes the default id;
- remaining props are spread as element attributes and `testid` is not duplicated as a `testid` attribute;
- the shared store mirrors received props (including non-serializable values like functions) and excludes `testid`;
- mounting more than one instance with distinct testids renders both, queryable individually, with per-instance props and slot content.

## Verification

- Full `ui-components` vitest suite: `640 passed`. The only red, `ShareChoicesButton > shows and hides copied message when clicked`, is a pre-existing flaky CSS-animation timing test unrelated to this change — it passes on an isolated rerun.
- Discrimination check: running the new `MockComponent.test.ts` against the original (unfixed) mock hangs on the multi-instance test — the exact #1840 symptom — confirming the test would catch a regression.


thedavidmeister:
Reviewed 020b79aeb: approved by maintainer; all required + sol checks green, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2730, opened 2026-06-13
Show trade amounts as a % of current vault balance on the quote table

### LINKED ISSUE #951: show trade amounts on the quote table as a % of current vault balances

thedavidmeister:
Fixes #951

## What the issue wanted

On the webapp quote table, show each trade's amount as a percentage of the **current vault balance** (alongside the absolute amount), so it's easy to visually spot trades that are large relative to their vaults — i.e. **drawdown management**.

## Output-only (revised)

Drawdown is depletion: it's the **output** vault you sell from that gets drawn down. The input side measures inflow/accumulation, not drawdown, so it was misleading. This PR now shows **only** the output percentage:

- **Maximum Output** → `maxOutput / output-vault balance * 100`

The percentage is computed in the binding layer (`crates/common`), because Float division/multiplication is only exposed on the Rust side, not in the TS Float API. It reuses the **already-fetched subgraph vault balances** — no extra network calls — and is attached to each quote's `data` as the optional field `formattedMaxOutputAsPercentOfVault`.

The output vault is matched to the quote pair by `(token, vaultId)` against the order's decoded on-chain `validOutputs` (the pair output index addresses those, and their ordering is independent of the subgraph `outputs` array — so an index-only match would be wrong). When the vault can't be matched or its balance is zero (percentage undefined / division by zero), the field is omitted and the UI shows the absolute amount alone.

The `formattedMaxInputAsPercentOfVault` field, its Rust computation, the `.d.ts`/wasm surface, the Svelte render, and the input-specific tests have all been **removed**.

## Where it renders (rendering fix)

The output percentage renders as a small secondary line directly under the **Maximum Output** amount in `TanstackOrderQuote` — the quote table on the order detail page (`OrderDetail` → `TanstackOrderQuote`, route `/orders/[chainId]-[raindex]-[orderHash]`). E.g. `25% of vault`.

The render and data path were verified end-to-end against a fresh build (not just a mocked unit test):
- The wasm-boundary test asserts the **exact** computed value (`maxOutput 1 / balance 10 * 100 = "10"`) reaches JS under the camelCase key the UI reads — proving the field survives the wasm/serde boundary (a `typeof === "string"` smoke check would not have caught a serialization regression).
- A component test asserts the percentage renders **inside the same cell** as the Maximum Output amount, and is absent when the field is `None`; mutation-validated (deleting the render block makes the test fail).
- The compiled deployable webapp bundle (`.svelte-kit/output/.../[orderHash]/+page.svelte.js` and the client chunk) contains the render and the `formattedMaxOutputAsPercentOfVault` field.

Note: the wasm bindings (`packages/raindex/dist`) and `ui-components/dist` are gitignored and regenerated on every build/deploy; the field appears in the running app once those are rebuilt, which the preview/prod Vercel build (`script/vercel-build.sh`: raindex → ui-components → webapp) does.

## Verification

- Rust `raindex_common` `order_quotes`: 7 tests green (output `"25"`/`"10"`; zero output-vault balance omits the percentage; mismatched-vault fixture yields `None`). Input-percentage tests removed.
- `@rainlanguage/raindex`: wasm builds, `tsc` check passes, generated `.d.ts` carries `formattedMaxOutputAsPercentOfVault?` only. Full JS suite: 124 pass.
- `@rainlanguage/ui-components`: full vitest suite 641 pass, `svelte-check` and eslint clean.
- webapp builds; the order-detail page bundle includes the render.


thedavidmeister:
screenshot

UI screenshot: Maximum Output now shows N% of the output vault balance (drawdown view), output-only.

thedavidmeister:
Reviewed a1526a8c2: approved by maintainer; CI green on all required + sol checks, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2733, opened 2026-06-13
Add integration tests for clearing/taking dead orders

### LINKED ISSUE #624: tests for clearing and taking dead orders

thedavidmeister:
Fixes #624 — adds integration-test coverage in `crates/integration_tests` for clearing and taking against dead/expired orders, asserting the best-effort no-op / `OrderNotFound` behaviour. Test-only; no contract changes.

## What the issue wanted
Tests for `clear3` and `takeOrders4` against dead orders. The contract treats a dead order as a best-effort no-op (emit `OrderNotFound`, skip/return early) rather than reverting, so a stale or removed order can't brick a bulk clear/take. There was no end-to-end coverage of this over a live `LocalEvm` — `crates/integration_tests` only had the two `post_task` tests.

## What this adds
Three `#[tokio::test]`s that drive the real deployed `RaindexV6` on a `LocalEvm`. An order is killed via `removeOrder3` (after which `orderExists` returns `false` / `sOrders[hash] == ORDER_DEAD`):

- **`test_take_orders_dead_order_is_noop`** — `takeOrders4` against a single dead order succeeds (no revert), emits exactly one `OrderNotFound { sender = caller, owner, orderHash }`, emits no `TakeOrderV3`, and returns zero `totalTakerInput`/`totalTakerOutput` (also asserted via a static `call`).
- **`test_take_orders_dead_order_skipped_live_order_taken`** — a `[dead, live]` batch: exactly one `OrderNotFound` (naming the dead order's hash) **and** exactly one `TakeOrderV3`; the live order's full `outputMax = 2` is filled at ratio 1 (taker input/output = 2). Proves the dead order is skipped while the live one is still taken.
- **`test_clear_dead_order_is_noop`** — `clear3` with one dead side succeeds, emits `OrderNotFound` for the dead owner/hash, and returns early so **neither** `ClearV3` **nor** `AfterClearV2` is emitted.

## Verification
- `cargo test -p raindex_integration_tests`: 5 passed, 0 failed (3 new + 2 pre-existing, all green; no env/RPC/deploy-dependent failures in this suite).
- `cargo fmt -- --check`: clean. No new warnings.
- **Discrimination check**: temporarily mutated each new test's expected values (OrderNotFound count `1→2`, taker input `2→3`, dead owner `alice→bob`) and confirmed all three FAIL with the real on-chain values in the diagnostic (`left: 1 / right: 2`, owner `0xf39f… vs 0x7099…`, etc.), then reverted. The pre-existing tests stayed green throughout. So the assertions catch a regression that reverts on, double-counts, or silently fills a dead order — not just "it runs".


thedavidmeister:
Reviewed cb2dca197: approved by maintainer; CI green on all required + sol checks, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2734, opened 2026-06-13
Mask signer<n>() row operand with & 0xFF to match signed-context (#2653)

thedavidmeister:
Split out of #2692 for independent review (issue #2653). Re-implemented on current main post-#2701. Note: testProdDeploy*/start-block/constant-suite tests are red pending the serial deploy at land time.


thedavidmeister:
Closing per review. The & 0xFF mask on subParserSigners isn't justified: unlike the sibling subParserSignedContext (which masks because its operand is a packed two-byte [column][row] field, so the mask *extracts* a byte), `signers` takes a single unbounded operand as the row. Masking it doesn't create parity — it silently converts an explicit out-of-range error into a wrong result: signer<256>() would resolve to signer<0>() instead of reverting ContextGridOverflow(5, 256). The current revert-on-out-of-range is the safer behavior. The #2653 'inconsistency' is by design (the two words differ because their operand encodings differ). If bounded/wrapping signer indices are ever desired, that should be explicit + parse-time validated, not a silent runtime mask. #2653 can be closed as by-design if you agree.

PR #2731, opened 2026-06-13
Conform to one-contract-per-.sol: extract inline test helpers

thedavidmeister:
## What

Conforms raindex to the **one-contract-per-`.sol`** convention so that `rainix-sol/static` goes green. The rainix #214/#225 check counts top-level `contract` + `abstract contract` declarations per tracked `.sol` and fails when > 1; the maintainer confirmed it applies to **test** files too, same as `src`. This PR conforms by **splitting** the offending files — it does not weaken the check.

19 tracked `.sol` files declared more than one top-level contract. Each inline helper (mocks, harnesses, stubs, child deployables, token doubles) is extracted **verbatim** into its own file under `test/util/concrete/` — the repo's existing home for standalone concrete test helpers (alongside `MockToken`, `ChildRaindexV6ArbOrderTaker`, etc.), one contract per file named after the contract. The only file with two *test* contracts (`RaindexV6ArbCommon.fallback.t.sol`) has its second test contract moved to its own sibling test file. Imports/references are rewired across the repo; no test logic changes.

## Files split (19) → contracts extracted

- `test/abstract/RaindexV6FlashLender.transfers.t.sol` → `TKN`, `IPull`, `Alice`, `Bob`, `Carol`
- `test/lib/LibGenericPoolExchange.spenderPool.t.sol` → `LibGenericPoolExchangeHarness`, `AllowanceProbePool`, `FallbackPool`
- `test/lib/LibGenericPoolExchange.zeroAddress.t.sol` → (now imports the shared `LibGenericPoolExchangeHarness`)
- `test/lib/LibRaindexArb.finalizeArbZeroGuard.t.sol` → `LibRaindexArbFinalizeHarness`, `RevertOnZeroReceive`
- `test/lib/LibRaindex.doPost.t.sol` → `LibRaindexDoPostHarness`
- `test/concrete/raindex/RaindexV6.clear.freshDriver.t.sol` → `CalcInterpreter`, `RaindexV6ClearDriverHarness`
- `test/concrete/raindex/RaindexV6.calculateOrderIOHarness.t.sol` → `FixedStackInterpreter`, `RaindexV6CalculateOrderIOHarness`
- `test/concrete/raindex/RaindexV6.clear.calculateStateChange.t.sol` → `RaindexV6ClearStateChangeHarness`
- `test/concrete/raindex/RaindexV6.clear.mock.t.sol` → `MockInterpreter`
[...]

PR #2738, opened 2026-06-14
fix(test): repair quote tests broken by the #2701 OrderNoSources guard

thedavidmeister:
## What

PR #2701 (issue #2652) added an `OrderNoSources()` source-count guard to
`addOrder4`: it reverts when an order's evaluable bytecode deserializes to
fewer than two sources (calculate + handle IO). #2701 updated most tests for
the new guard but missed several that construct **source-less** orders, so
those tests now revert during their setup — before reaching the behavior they
assert.

This PR repairs them. **Test-only**, no source/artifact/pointer changes.

### The three quote tests (the assigned breakage)

In `test/concrete/raindex/RaindexV6.calculateOrderIOHarness.t.sol`, three
`quote2` tests add a live order via `addOrder4` with empty (`hex""`) bytecode,
which now reverts `OrderNoSources()` before the quote assertion:

- `testQuoteLiveOrderReturnsCappedMaxAndRatio`
- `testQuoteLiveSelfTradeReverts`
- `testQuoteRemovedOrderReturnsFalseZeroZero`

Fix: give the harness order the canonical two-source
`LibTestAddOrder.VALID_ORDER_BYTECODE` (the same `[constants][bytecode]` blob
the passing add/quote suites use). The harness drives a fixed-stack
interpreter that ignores the bytecode, so `calculateOrderIO`/`quote2` behave
identically and each test still asserts exactly what it did before
(vault-capped max + ratio; self-trade revert; removed-order `(false, 0, 0)`).

### Two more tests broken by the same #2701 guard

While running the full suite to verify, the same guard was found breaking five
`addOrder4` tests in `test/concrete/raindex/RaindexV6.orderMgmt.fresh.t.sol`,
which set a 13-byte `CALC_BYTECODE` blob that deserializes to **zero** sources.
They hit `OrderNoSources()` instead of the input/output/meta guard each one
targets (`testFreshAddOrderNoInputsReverts`, `testFreshAddOrderOneInputPasses`,
`testFreshAddOrderNoOutputsReverts`,
`testFreshAddOrderNoInputsTakesPrecedenceOverNoOutputs`,
`testFreshAddOrderMetaEmitsExact`). Fix: point `CALC_BYTECODE` at
`VALID_ORDER_BYTECODE` and correct its now-stale "any bytecode is accepted"
comment.

### Compile blocker that masked all of the above

`test/concrete/parser/RaindexV6SubParser.routingTokenDecimalsAndMasks.t.sol`
imported `rainlang-0.1.2/...` while the repo pins `rainlang-0.1.5`. The path
didn't resolve, so `forge test` failed to **compile** the whole suite — which
is why `rainix-sol/test` is currently red on `main` with a `Source ... not
found` compiler error rather than test failures. The `OrderNoSources` failures
above were latent behind it. Fix: bump the two imports to `rainlang-0.1.5`
(both symbols exist there and `src/lib/LibRaindexSubParser.sol` already imports
`LibSubParse` from `rainlang-0.1.5`).

## Verification

`nix develop github:rainlanguage/rainix#sol-shell -c forge test`:
**675 passed, 12 failed, 1 skipped** (was 670/17 with the compile blocker
hand-patched). The 8 quote/orderMgmt tests above all pass. `forge fmt --check`
is clean.

The 12 remaining failures are pre-existing and **environment-only**, unrelated
to this change:
- `testProdDeploy{Arbitrum,Base,BaseSepolia,Flare,Polygon}`,
  `testStartBlock{Base,Flare,Polygon}`,
  `testIsStartBlock{Base,Flare,Polygon}` — `vm.createSelectFork: environment
  variable '<CHAIN>_RPC_URL' not found` (need per-chain RPC endpoints).
- `testSubgraphYamlAddress` — `vm.ffi` can't find `yq` on the local PATH.

## Summary by CodeRabbit

* **Tests**
  * Updated test dependencies and improved bytecode initialization in order management tests for enhanced validation coverage.

### DIFF
--- a/test/concrete/parser/RaindexV6SubParser.routingTokenDecimalsAndMasks.t.sol
+++ b/test/concrete/parser/RaindexV6SubParser.routingTokenDecimalsAndMasks.t.sol
@@ -5,8 +5,8 @@ pragma solidity =0.8.25;
 import {Test} from "forge-std-1.16.1/src/Test.sol";
 import {OperandV2} from "rain-interpreter-interface-0.1.0/src/interface/ISubParserV4.sol";
 import {OPCODE_CONTEXT} from "rain-interpreter-interface-0.1.0/src/interface/IInterpreterV4.sol";
-import {ContextGridOverflow} from "rainlang-0.1.2/src/error/ErrSubParse.sol";
-import {LibSubParse} from "rainlang-0.1.2/src/lib/parse/LibSubParse.sol";
+import {ContextGridOverflow} from "rainlang-0.1.5/src/error/ErrSubParse.sol";
+import {LibSubParse} from "rainlang-0.1.5/src/lib/parse/LibSubParse.sol";
 import {LibRaindexSubParser} from "../../../src/lib/LibRaindexSubParser.sol";
 
 /// @title RaindexV6SubParserRoutingTokenDecimalsAndMasksTest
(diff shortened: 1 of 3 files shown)


thedavidmeister:
Reviewed 0bfce9aed: repairs all 8 OrderNoSources test regressions from #2701 (3 quote + 5 orderMgmt; source-less test orders -> valid 2-source bytecode) -> rainix-sol/test now GREEN. The only red is rainix-sol/static, a stale-behind-base artifact (this branch predates #2731's one-contract split); the PR is MERGEABLE (no conflict) and test-only, so merging into current main — which already has the split — keeps static green while greening test. This clears main's last red. LGTM.

PR #2740, opened 2026-06-14
Fix selected chain ids store pruning in webapp

### LINKED ISSUE #2208: Fix selected chain ids store variable in webapp
The store incorrectly persist the data that is not found in the setting configuration. Sometimes there are stale data left from another settings applied before so this interferes with the orders list and vaults list queries.

We need to make sure the selected chain ids store only keeps the current settings content and remove any unknown chain ids from local storage.

thedavidmeister:
## Problem

Closes #2208.

`selectedChainIds` (a `cachedWritableStore` persisted to local storage) is meant to only keep chain IDs that belong to the current settings configuration. A `validChainIds.subscribe(...)` block in `packages/webapp/src/lib/stores/settings.ts` is responsible for pruning any selected IDs that aren't in the currently-valid set, so stale IDs left over from a previously-applied settings don't leak into the orders/vaults list queries.

That pruning didn't reliably work. The subscriber read the current selection with `get(selectedChainIds)`:

```ts
validChainIds.subscribe((valid) => {
	if (valid.length > 0) {
		const current = get(selectedChainIds);
		const filtered = current.filter((id) => valid.includes(id));
		if (filtered.length !== current.length) {
			selectedChainIds.set(filtered);
		}
	}
});
```

Under Svelte's store subscriber-queue re-entrancy, calling `get()` (a transient subscribe) inside another store's subscriber while a flush is still draining can deliver `undefined` instead of the live value. In that case `current.filter(...)` either throws `Cannot read properties of undefined (reading 'filter')` or reads a stale value and silently skips the prune — so stale chain IDs survive in `selectedChainIds` and in local storage. That is exactly the behavior #2208 reports ("the store incorrectly persists data that is not found in the setting configuration ... this interferes with the orders list and vaults list queries").

## Fix

Read the current selection atomically via `selectedChainIds.update((current) => ...)` instead of a nested `get()`. `update` reads the writable's live internal value with no transient subscribe, so it is immune to the re-entrancy hazard. The no-op short-circuit is preserved by returning the existing array unchanged when nothing was pruned:

```ts
validChainIds.subscribe((valid) => {
	if (valid.length > 0) {
		selectedChainIds.update((current) => {
			const filtered = current.filter((id) => valid.includes(id));
			return filtered.length !== current.length ? filtered : current;
		});
	}
});
```

`get` is no longer imported.

## Tests

New `packages/webapp/src/lib/stores/settings.test.ts` (Vitest + jsdom, same harness as `wagmi.test.ts`) pins the behavior #2208 asks for:

- stale IDs absent from `validChainIds` are pruned once `validChainIds` is populated;
- valid IDs are retained;
- an empty `validChainIds` leaves the selection untouched (the `length > 0` guard — important so a fresh load doesn't wipe the persisted selection before the app populates `validChainIds`);
- re-pruning when `validChainIds` later narrows;
- the pruned (not the stale) selection is what serializes / round-trips through local storage.

Each test is mutation-validated — it provably discriminates:

| Mutation to source | Tests that fail |
| --- | --- |
| invert filter predicate (`valid.includes` → `!valid.includes`) | 5 (all prune + persisted-prune) |
| weaken empty guard (`valid.length > 0` → `>= 0`) | 1 ("leaves untouched while empty") |
| drop the write-back (always return `current`) | 4 (prune + persisted-prune) |
| revert to the original buggy `get()`-based implementation | 6 |

Full `packages/webapp` Vitest suite is green (26 files / 188 tests) with these changes; `eslint` is clean on both files.

No bytecode / ABI / artifact changes — this is a pure webapp TypeScript fix, so no redeploy or constant re-pin is needed.


thedavidmeister:
human-approved; merging

PR #2741, opened 2026-06-14
Show Syncing in top-level localdb status badge

### LINKED ISSUE #2565: top-level localdb status should show Syncing in addition to Active / Failure
## Problem

The top-level \`LOCALDB\` indicator in the webapp sidebar collapses many per-chain states into one of two visible labels (\`ACTIVE\` or \`FAILURE\`). When chains are partway through their pipeline — manifest fetched, dumps downloading, engine running, etc. — the per-chain detail modal shows \`SYNCING\` but the sidebar badge does not reflect that state.

This means a user looking at the sidebar can't tell the difference between:

- All chains are healthy and idle (\`ACTIVE\`)
- All chains are healthy and busy bootstrapping / fetching / decoding (also surfaces as something other than \`FAILURE\` today)

If even one chain is mid-pipeline, the top-level badge should show \`SYNCING\` so the user knows the database isn't authoritative yet and will change as work completes.

## Proposal

Add a \`SYNCING\` state to the top-level localdb status badge with priority below \`FAILURE\` and above \`ACTIVE\`:

- If any chain is in \`FAILURE\` → top-level badge shows \`FAILURE\` (existing behaviour).
- Else if any chain is in \`SYNCING\` (anything other than the steady-state idle/ready phase) → top-level badge shows \`SYNCING\`.
- Else → top-level badge shows \`ACTIVE\`.

## Acceptance

- During an in-progress sync (e.g. immediately after registry refresh, dumps downloading, etc.), the sidebar \`LOCALDB\` badge reads \`SYNCING\` rather than \`ACTIVE\`.
- Once every chain reaches the post-sync ready state, the badge transitions to \`ACTIVE\`.
- When any chain fails, the badge shows \`FAILURE\` regardless of the others' progress.

thedavidmeister:
## Problem

The top-level `LOCALDB` badge in the webapp sidebar only ever shows `ACTIVE` or `FAILURE`. While a chain is partway through its pipeline (manifest fetched, dumps downloading, engine running, etc.) the per-chain modal correctly reads `SYNCING`, but the sidebar badge does not — so the user can't tell a healthy-and-idle database from one that is busy bootstrapping and not yet authoritative.

## Root cause

The sidebar renders `LocalDbStatusCard` directly (`Sidebar.svelte`), and the card computed its badge as:

```ts
$: hasFailure = networkList.some((s) => s.status === 'failure');
$: displayStatus = (hasFailure ? 'failure' : 'active') as LocalDbStatus;
```

This only inspects `networkStatuses`, never emits `'syncing'`, and ignores `raindexStatuses` entirely. The `LocalDbStatusBadge` already renders a `syncing` state, and the webapp `aggregateStatus` store already implemented the correct `failure > syncing > active` priority across both lists — but the sidebar badge used the card's own (buggy) logic, not the store.

## Fix

- Add a shared `aggregateLocalDbStatus(statuses)` helper in `ui-components` that collapses a combined list of per-chain / per-raindex statuses into a single top-level status using the priority **`failure` > `syncing` > `active`** (empty input ⇒ `active`). This mirrors the canonical Rust rule in `crates/common/.../local_db/status.rs`.
- The card now derives `displayStatus` from this helper over `networkStatuses` **and** `raindexStatuses`.
- The webapp `aggregateStatus` store (which had the same rule hand-rolled) now calls the same helper, so the card and the store can't drift.

Per the issue's acceptance criteria: a mid-sync chain ⇒ badge reads `Syncing`; all chains ready ⇒ `Active`; any failure ⇒ `Failure` regardless of others' progress.

## How the tests discriminate (mutation-validated)

Updated `LocalDbStatusCard.test.ts`. The previous case "shows active status when all networks are syncing" asserted `Active` (enshrining the bug) — it now asserts `Syncing`, per the issue. Added cases for a syncing-only network, a syncing-only raindex, and `failure`-beats-`syncing` precedence (network failure and raindex failure, each with a sibling syncing).

Mutation testing against the helper:
- Mutating the syncing branch to return `active` instead of `syncing` ⇒ exactly the 3 syncing assertions fail (network-syncing, all-syncing, raindex-syncing).
- Reordering so syncing is checked before failure ⇒ exactly the 2 `failure`-beats-`syncing` precedence assertions fail.
Restoring the source returns the suite to green, so the tests provably pin both the syncing emission and the `failure > syncing` priority.

## Verification

- Full `@rainlanguage/ui-components` suite: 640 tests pass.
- Full `@rainlanguage/webapp` suite: 180 tests pass (store's existing 26 `localDbStatus` tests still pass after the refactor to the shared helper).
- `ui-components` prettier + eslint + `svelte-check`: clean (0 errors / 0 warnings). `webapp` eslint: clean.
- The only `webapp` `svelte-check` reds are the pre-existing `PUBLIC_WALLETCONNECT_PROJECT_ID` env-var errors in `handleWalletInitialization.ts` (unset `.env`), unrelated to this change.

Frontend/TS only — no Solidity, no Rust source, no generated artifacts or bytecode changes; no redeploy needed.

Closes #2565

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for status card to validate additional sync and failure state combinations.

* **Refactor**
  * Centralized local database status aggregation logic with improved state priority handling (failure overrides syncing, syncing overrides active).

### DIFF
--- a/packages/ui-components/src/__tests__/LocalDbStatusCard.test.ts
+++ b/packages/ui-components/src/__tests__/LocalDbStatusCard.test.ts
@@ -29,10 +29,24 @@ describe("LocalDbStatusCard", () => {
     expect(screen.getByTestId("local-db-status")).toBeInTheDocument();
   });
 
-  it("shows active status when no networks have failures", () => {
+  it("shows syncing status when a network is syncing and none are failing", () => {
     const networkStatuses = new Map<number, NetworkSyncStatus>([
       [1, { chainId: 1, status: "active", schedulerState: "leader" }],
       [137, { chainId: 137, status: "syncing", schedulerState: "leader" }],
+    ]);
+
+    render(LocalDbStatusCard, {
+      props: { networkStatuses },
+    });
+
+    expect(screen.getByText("LocalDB")).toBeInTheDocument();
+    expect(screen.getByText("Syncing")).toBeInTheDocument();
+  });
+
+  it("shows active status when all networks are active", () => {
+    const networkStatuses = new Map<number, NetworkSyncStatus>([
+      [1, { chainId: 1, status: "active", schedulerState: "leader" }],
+      [137, { chainId: 137, status: "active", schedulerState: "leader" }],
     ]);
 
     render(LocalDbStatusCard, {
@@ -77,7 +91,7 @@ describe("LocalDbStatusCard", () => {
     expect(screen.getByTestId("local-db-status-header")).toBeInTheDocument();
   });
 
-  it("shows active status when all networks are syncing but none failing", () => {
+  it("shows syncing status when all networks are syncing but none failing", () => {
     const networkStatuses = new Map<number, NetworkSyncStatus>([
       [137, { chainId: 137, status: "syncing", schedulerState: "leader" }],
       [42161, { chainId: 42161, status: "syncing", schedulerState: "leader" }],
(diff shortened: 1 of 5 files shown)


thedavidmeister:
screenshot

UI screenshot: "Syncing" shown in the top-level localdb status badge

thedavidmeister:
Reviewed 6f8549a7c: approved by maintainer; all required + sol checks green, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2745, opened 2026-06-14
fix(ui-components): show actual order trade count instead of >99

### LINKED ISSUE #1992: Orders page - 3 enhancements
1. Show actual trades instead of >99 if trade count is >99
2. Order / sort by columns 
3. If input or output has >2 tokens, wrap the row

Image

thedavidmeister:
## What

Part 1 of #1992 (Orders page enhancements): show the **actual** trade count in the Orders table instead of the `>99` placeholder.

`packages/ui-components/src/lib/components/tables/OrdersListTable.svelte` rendered the trades cell as:

```svelte
{item.tradesCount > 99 ? '>99' : item.tradesCount}
```

The underlying `trade_count` comes from a full SQL `COUNT(*)` (`crates/common/src/local_db/query/fetch_order_trades_count/query.sql`) and is never capped upstream, so the `>99` cap only hid real, already-computed information from the user. The cell now renders the real number:

```svelte
{item.tradesCount}
```

## Scope

Deliberately limited to **part 1** of #1992 (the `>99` display). Column sorting (part 2) and >2-token row-wrapping (part 3) are intentionally left for separate PRs.

## How the test discriminates

`packages/ui-components/src/__tests__/OrdersListTable.test.ts` already had a `handles large number of trades display` case rendering an order with `tradesCount: 100`. Its expectation was changed from `>99` to the exact `100`. Mutation-validated: restoring the old `> 99 ? '>99' : ...` ternary makes that test fail (`Expected: 100 / Received: >99`), and restoring the fix makes all 13 tests in the file pass — so the test discriminates the behavior, not just the wiring.

## Deploy dependency

None — this is a pure front-end rendering change with no contract/bytecode change.

Closes #1992

### DIFF
--- a/packages/ui-components/src/__tests__/OrdersListTable.test.ts
+++ b/packages/ui-components/src/__tests__/OrdersListTable.test.ts
@@ -492,7 +492,7 @@ describe("OrdersListTable", () => {
     })) as Mock;
 
     render(OrdersListTable, defaultProps as OrdersListTableProps);
-    expect(screen.getByTestId("orderListRowTrades")).toHaveTextContent(">99");
+    expect(screen.getByTestId("orderListRowTrades")).toHaveTextContent("100");
   });
 
   it("shows Take button for active orders", async () => {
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed 46d5966a9: small ui-components fix — shows the actual order trade count instead of capping the display at >99. Any rs/wasm/sol reds are the stale-base + cachix-action infra flake (main is green on them; rainix #226 pins cachix), not from this UI change. LGTM.

thedavidmeister:
screenshot

UI screenshot: Orders list "Trades" column now shows the actual count instead of the old >99 cap.

PR #2749, opened 2026-06-14
Fix red main: split arb test into one-contract-per-file

thedavidmeister:
## Problem

`main` is red on the shared `rainix-sol / static` job. The
`rainix-sol-single-contract` check fails:

```
ERROR: test/concrete/arb/RouteProcessorRaindexV6ArbOrderTaker.noEthForward.t.sol declares 2 contracts; Rain convention is one contract per file.
```

The file declared two top-level contracts: the `ValueRecordingRouteProcessor`
mock and the `RouteProcessorRaindexV6ArbOrderTakerNoEthForwardTest` test. This
reds `rainix-sol / static` on main and every open PR rather than being a per-PR
cause.

## Fix

Move the `ValueRecordingRouteProcessor` mock (verbatim body) to
`test/util/concrete/ValueRecordingRouteProcessor.sol` — matching the existing
`test/util/concrete/` mock convention (e.g. `MockRouteProcessor.sol`) — and
import it. The test file now declares exactly one contract. The now-unused
`IRouteProcessor` import (only the moved mock used it) is removed from the test
file.

## Local proof

In `nix develop github:rainlanguage/rainix#sol-shell`:

- `rainix-sol-single-contract` -> exit 0 (was exit 1)
- `forge fmt --check` on both files -> clean
- `forge test --match-path test/concrete/arb/RouteProcessorRaindexV6ArbOrderTaker.noEthForward.t.sol`
  -> `1 passed; 0 failed` (`testProcessRouteCalledWithNoEthValue`, 5096 runs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Refactored test utilities by consolidating mock route processor implementations to reduce duplication.

PR #2756, opened 2026-06-15
test: mutation-harden batch_quote coverage

thedavidmeister:
## Module hardened

`crates/quote/src/rpc.rs` — the **batch-quote assembly** core of order quoting: partitioning quote targets by `raindex` address into separate `multicall(bytes[])` groups, chunking each group by `chunk_size`, the OZ-Multicall revert **bisection** that isolates a reverting target, and scattering all per-group / per-chunk / per-bisection results back into the caller's input order. This is correctness-critical: a mis-routed quote feeds a wrong max-output/io-ratio to the bot/UI.

Test-only PR. No production code changed (433 insertions, 0 deletions, all inside `#[cfg(test)] mod tests`).

## Mutation matrix

Each behavior was validated by mutating the exact source line, confirming the new test fails (kills the mutant), then restoring. **7/7 mutants killed.**

| # | Mutated line | Mutation | Killing test | Verdict |
|---|---|---|---|---|
| M1 | `groups…find(\|(ob,_)\| *ob == target.raindex)` | `==` → `!=` | `test_batch_quote_groups_by_raindex_and_scatters_in_input_order` | KILLED |
| M2 | `results[idx] = Some(r)` (scatter) | `idx` → `0` | `…groups_by_raindex…` | KILLED |
| M3 | `(chunk_start + chunk_size).min(len)` | `+` → `*` | `test_batch_quote_chunks_group_and_preserves_order` | KILLED |
| M4 | `out.push((indexes[chunk_start + offset], …))` | drop `chunk_start` | `…chunks_group…` | KILLED |
| M5 | `if decoded.len() != quote_targets.len()` (length guard) | `!=` → `==` | `test_batch_quote_multicall_length_mismatch_is_rejected` | KILLED |
| M6 | `if chunk.len() == 1` (bisection base case) | `== 1` → `== 2` | `test_batch_quote_mixed_chunk_bisects_success_and_revert_by_offset` | KILLED |
| M7 | `resolved[start + offset] = Some(result)` (bisection OK branch) | drop `start` | `test_batch_quote_bisection_multi_element_ok_subchunk_lands_at_offset` | KILLED |

The 11 pre-existing `rpc::tests` were audited as part of the campaign: they validate the happy path, non-existent orders, registry-decoded / unknown-selector reverts, counterparty threading, and uniform-revert bisection — but left the grouping/chunking/scatter arithmetic, the length-mismatch guard, and the mixed-success/revert bisection offsets uncovered. The five new tests close exactly those gaps.

## New tests

- `test_batch_quote_groups_by_raindex_and_scatters_in_input_order` — two distinct raindex addresses fall into separate groups; each group's quote scatters back to its original input slot (kills M1, M2).
- `test_batch_quote_chunks_group_and_preserves_order` — `chunk_size = 2` over 3 same-raindex targets splits into `[0,1]` + `[2]`; results reassemble in input order, each chunk's returned `bytes[]` length matching the chunk (kills M3, M4).
- `test_batch_quote_multicall_length_mismatch_is_rejected` — a multicall returning more elements than targets is rejected as `CorruptReturnData` carrying the mismatch detail, never silently mis-aligned (kills M5).
- `test_batch_quote_mixed_chunk_bisects_success_and_revert_by_offset` — a 2-target chunk where one quotes and one reverts bisects to singletons, placing the success and the decoded revert at their correct offsets (kills M6).
- `test_batch_quote_bisection_multi_element_ok_subchunk_lands_at_offset` — a 4-target chunk (reverting pair + quoting pair) whose `[2,4)` sub-chunk succeeds wholesale: the two successes land at slots 2/3 via `resolved[start + offset]`, not clobbering the reverting pair's slots (kills M7).

## Gaps checklist

- [x] raindex grouping condition
- [x] per-group result scatter into input order
- [x] chunk boundary stepping / `.min` clamp
- [x] per-chunk reassembly offset
- [x] multicall length-mismatch guard
- [x] bisection singleton base case
- [x] bisection OK-branch offset placement (incl. multi-element OK sub-chunk)
- [ ] live RPC transport-error path (`TransportError`) — exercised only via integration; left as-is (RPC glue, out of scope for pure-logic hardening)

## Real bugs

None found. The batch-assembly logic is correct under every mutation probed.

## Verify

- `nix develop -c cargo test -p raindex_quote` → **56 passed, 0 failed, 0 ignored** locally.
- `cargo fmt -p raindex_quote -- --check` clean; `cargo clippy -p raindex_quote --all-targets` clean.

## Inherited-red note

raindex `main` is currently red on the 4-fix-set causes (rs-static rustfmt, sol-static, sol-test, wasm). A Rust coverage PR branched off `main` **inherits** those reds (especially `rs-static`) until #2749–2752 merge — those failures are pre-existing and **not** introduced by this PR, which only adds tests to one crate.


thedavidmeister:
human-approved; merging

PR #2752, opened 2026-06-14
Fix red main: gate allowance.rs tests to non-wasm (httpmock)

thedavidmeister:
## Problem

`main` is red on `wasm-test` and `wasm-browser-test` (both required checks). The
`raindex_common` lib tests fail to compile for `wasm32-unknown-unknown`:

```
error[E0432]: unresolved import `httpmock`
  --> crates/common/src/allowance.rs:26:9
26 |     use httpmock::MockServer;
```

`allowance.rs`'s `mod tests` was gated only `#[cfg(test)]`, but it uses
`httpmock` (a `[target.'cfg(not(target_family = "wasm"))'.dev-dependencies]`
dep) and `tokio::runtime::Runtime`, so it compiled under wasm where httpmock is
absent. The other 6 `error[E0282]: type annotations needed` were knock-on
effects of the failed crate compile, not independent problems. This reds both
wasm jobs on main and every open PR.

## Fix

Gate the module `#[cfg(all(test, not(target_family = "wasm")))]`, matching every
other httpmock-using test module in the crate (`erc20.rs`, `rpc_client.rs`,
`dotrain_order.rs`, vaults' `mod non_wasm`, ...). The tests still run on native
(`rs-test`).

## Local proof

In `nix develop .#wasm-shell`:

- `CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test --no-run --target wasm32-unknown-unknown --lib -p raindex_common`
  -> `Finished`, exit 0 (was exit 101, `error[E0432]: unresolved import httpmock`)
- `cargo test --no-run -p raindex_common --lib` (native) -> `Finished`, exit 0
  (tests still compile on native).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2759, opened 2026-06-15
test: mutation-harden SgVaultBalanceChangeType accessor coverage

thedavidmeister:
## Module

`crates/subgraph/src/types/common.rs` — the `SgVaultBalanceChangeType` accessor methods (`typename`, `timestamp`, `amount`, `new_vault_balance`, `transaction`). These are pure response→domain mapping: each method `match`es the enum's five variants (`Withdrawal`, `TradeVaultBalanceChange`, `Deposit`, `ClearBounty`, `Unknown`) and returns the corresponding field of the inner struct.

This is the weakest high-value pure-logic module in the subgraph client: `types/common.rs` had **no test module at all**, and these accessors were only exercised indirectly and partially:
- `typename()` — only via the `Deposit` variant (in `raindex_client/vault.rs` type-filter tests).
- `timestamp()` — only via `Deposit`/`Withdrawal` in the `vault_balance_changes_query.rs` sort test.
- `amount()`, `new_vault_balance()`, `transaction()` — **never asserted anywhere** in the crate.

By contrast, the query/filter-construction modules I surveyed (`raindex_client/order.rs::orders_list`, `raindex_client/vault.rs::vaults_list`, `pagination.rs`, `utils/slice_list.rs`) are already mutation-saturated with `body_contains` where-clause assertions, so this enum-mapping logic was the genuine gap.

## Mutation matrix

For each accessor arm I mutated the exact return expression (swap a variant's field to a sibling field, or `Some↔None`), ran the crate test suite, and recorded survival. On the unmutated suite **every** mutant below SURVIVED before this PR. After adding the discriminating tests, all are KILLED:

| Accessor | Withdrawal | TradeVaultBalanceChange | Deposit | ClearBounty | Unknown |
|---|---|---|---|---|---|
| `typename` | KILLED | KILLED | KILLED | KILLED | KILLED (literal) |
| `timestamp` | KILLED | KILLED | KILLED | KILLED | covered (`is_none`) |
| `amount` | KILLED | KILLED | KILLED | KILLED | covered (`is_none`) |
| `new_vault_balance` | KILLED | KILLED | KILLED | KILLED | covered (`is_none`) |
| `transaction` | KILLED | KILLED | KILLED | KILLED | covered (`is_none`) |

21 arm mutations covered, **0 survivors**. Each test builds one instance of every variant with field-distinct values (so an arm returning the wrong field is caught) and variant-distinct values (so an arm delegating to the wrong variant's struct is caught); the `Unknown` arm is pinned with `is_none()` / literal assertions.

## Gaps / notes

- One mutation in my harness (`timestamp` `Unknown → Some(&temp)`) is a build error (borrow of a temporary), not a coverage gap — the `Unknown` arm is already pinned by the `is_none()` assertion.
- No production code changed; diff is a single additive `#[cfg(test)] mod` (+163 lines). `cargo fmt` and `cargo clippy --tests` are clean; crate `cargo test` is green locally (185 passed, was 180).

## Inherited-red note

`main` is currently RED on unrelated 4-fix-set causes (rs-static rustfmt, sol-static, sol-test, wasm; tracked by #2749–#2752). A Rust coverage PR branched off `main` inherits those reds until they merge — they are **not** introduced by this change. The subgraph crate's own `cargo test` passes locally.

### DIFF
--- a/crates/subgraph/src/types/common.rs
+++ b/crates/subgraph/src/types/common.rs
@@ -933,3 +933,166 @@ mod impls {
     impl_wasm_traits!(SgTradeEventTypename);
     impl_wasm_traits!(SgTradeRef);
 }
+
+#[cfg(test)]
+mod sg_vault_balance_change_type_tests {
+    use super::*;
+
+    fn vault_ref() -> SgVaultBalanceChangeVault {
+        SgVaultBalanceChangeVault {
+            id: SgBytes("0xvault".to_string()),
+            vault_id: SgBytes("vault-id".to_string()),
+            token: SgErc20 {
+                id: SgBytes("0xtoken".to_string()),
+                address: SgBytes("0xtoken".to_string()),
+                name: Some("Token".to_string()),
+                symbol: Some("TKN".to_string()),
+                decimals: Some(SgBigInt("18".to_string())),
+            },
+        }
+    }
+
+    fn raindex_ref() -> SgRaindex {
+        SgRaindex {
+            id: SgBytes("0xraindex".to_string()),
+        }
+    }
+
+    // Each variant is built so that, within a variant, every accessor-exposed
+    // field carries a DISTINCT value (so an arm that returns the wrong field is
+    // caught), and so that the same accessor returns a DIFFERENT value across
+    // variants (so an arm that delegates to the wrong variant's struct is
+    // caught). The prefix encodes the variant; the suffix encodes the field.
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 94c8a56e5: approved by maintainer; all required + sol checks green, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2762, opened 2026-06-15
test: mutation-harden webapp deploy <+layout> coverage

thedavidmeister:
## Module

`packages/webapp/src/routes/deploy/+layout.ts` — the **deploy route's root `+layout.ts` `load` function**. It was entirely untested: no inline `import.meta.vitest` block and no sibling test file, unlike its `[orderName]` and `[deploymentKey]` child layouts which each ship a `layout.test.ts`. This was the weakest high-value untested webapp pure-logic module (route load logic with several `??` fallbacks and a catch ternary). Adds `packages/webapp/src/routes/deploy/layout.test.ts` (12 tests).

This is a standalone, **test-only** coverage PR. `git diff` touches no `src/` production code — only the new test file.

## Mutation matrix

Every behavior was validated by mutating the exact source line and running a fresh `npx vitest run src/routes/deploy/layout.test.ts` in isolation (no watch/cache reuse). All 13 mutants were **killed**, then the source was restored:

| # | Mutation | Behavior pinned | Result |
|---|---|---|---|
| M1 | drop `?? null` on `parentData.registry ?? null` | registry null-coalesce | KILLED |
| M2 | `if (!registry)` → `if (registry)` | registry-not-loaded guard | KILLED (12/12) |
| M3 | corrupt `'Registry not loaded'` string | guard error message | KILLED |
| M4 | `if (orderDetails.error)` → `if (!...)` | error early-return guard | KILLED |
| M5 | drop `readableMsg` in `readableMsg ?? msg` (registry error) | preferred message | KILLED |
| M6 | drop `?? msg` fallback (registry error) | fallback message | KILLED |
| M7 | drop `?? ''` on `registry.orders.get(name) ?? ''` | dotrain empty fallback | KILLED |
| M7b | wrong dotrain fallback value | dotrain empty fallback value | KILLED |
| M8 | drop `readableMsg` in invalid-order `readableMsg ?? msg` | preferred message | KILLED |
| M8b | drop `?? msg` fallback (invalid order) | fallback message | KILLED |
| M9 | success `error: null` → non-null | success return shape | KILLED (4 fail) |
| M10 | catch ternary → always `'Unknown error occurred'` | Error branch | KILLED |
| M10b | catch ternary → always `error.message` | non-Error branch | KILLED |

The 12 tests pass on the unmutated baseline (`12 passed`), pass within the full `npm run test -w @rainlanguage/webapp` suite, and eslint is clean on the new file.

## Gaps / saturation

None outstanding. The module was previously 0%-covered and is now branch-complete for `load()`. No real bug surfaced — the code is correct, just untested.

## Build / test recipe

Per `.github/workflows/test-webapp.yaml`, run inside `nix develop .#wasm-shell`: `npm install --no-check` → `npm run build -w @rainlanguage/raindex` (wasm) → `npm run build -w @rainlanguage/ui-components` → then vitest. Without the wasm + ui-components builds, vitest reds with "Failed to resolve entry".

## Inherited-red note

This branch is cut off `main`, which is currently RED on the 4-fix-set causes (rs-static, sol-static, sol-test, wasm) pending #2749–#2752. A coverage PR off main **inherits those reds** — they are pre-existing and **not** caused by this test-only change (it adds a single `*.test.ts` and edits no source). The relevant signal here is the webapp `test` job.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for the deploy layout functionality, validating error handling, order processing, and data structure integrity across various scenarios.

### DIFF
--- a/packages/webapp/src/routes/deploy/layout.test.ts
+++ b/packages/webapp/src/routes/deploy/layout.test.ts
@@ -1,0 +1,231 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { load } from './+layout';
+
+describe('deploy /+layout load function', () => {
+	const mockGetAllOrderDetails = vi.fn();
+	const mockOrdersGet = vi.fn();
+
+	const makeRegistry = () => ({
+		getAllOrderDetails: mockGetAllOrderDetails,
+		orders: { get: mockOrdersGet }
+	});
+
+	const mockParent = vi.fn();
+
+	beforeEach(() => {
+		vi.resetAllMocks();
+	});
+
+	const callLoad = () =>
+		load({
+			parent: mockParent
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		} as any);
+
+	it('returns "Registry not loaded" when parent registry is null', async () => {
+		mockParent.mockResolvedValue({ registry: null });
+
+		const result = await callLoad();
+
+		expect(result).toEqual({
+			validOrders: [],
+			invalidOrders: [],
+			registry: null,
+			error: 'Registry not loaded'
+		});
+		// getAllOrderDetails must not be reached when registry is missing
+		expect(mockGetAllOrderDetails).not.toHaveBeenCalled();
+	});
+
+	it('returns "Registry not loaded" when parent does not provide a registry', async () => {
+		// registry is undefined -> `?? null` coalescing path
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed b00b8d91a: approved by maintainer; all required + sol checks green, mergeable, CodeRabbit clean, comments current-behavior-only. Merging.

PR #2763, opened 2026-06-15
test: mutation-harden ui-components historicalOrderCharts coverage

thedavidmeister:
## Module

`packages/ui-components/src/lib/services/historicalOrderCharts.ts` — `prepareHistoricalOrderChartData`, which derives the value/time/color points for the historical order (io-ratio) chart from a list of trades.

A scoped adversarial-mutation-test coverage pass on this pure-logic derivation module. The two existing in-source tests covered the happy path (transform + sort, and weighted-average merge of same-timestamp trades) but left three behaviors unverified. Each new test was written to pass on clean code and fail under a re-applied mutation; all three are tests-only additions inside the existing `import.meta.vitest` block — production logic is untouched.

## Mutation matrix

| Behavior | Mutation | Before | After |
|---|---|---|---|
| Light-theme colour (`colorTheme != "dark"` -> `"#4E4AF6"`) | `"#4E4AF6"` -> `"#000000"` | **SURVIVED** | KILLED |
| `Math.abs` of the io ratio | drop `Math.abs(...)` | **SURVIVED** | KILLED |
| Single-trade branch threshold | `length > 1` -> `length > 0` | **SURVIVED** | KILLED |
| Weighted-average weight | `value * outputAmount` -> `value + outputAmount` | KILLED (existing) | KILLED |
| Final ordering | drop `sortBy(..., d => d.time)` | KILLED (existing) | KILLED |

Module was **not** mutation-saturated: 3 of 5 probed behaviors had surviving mutants, all now killed.

## Gaps closed

1. **Light theme colour.** Both existing tests only ever passed `"dark"`, so the `else` arm of the colour ternary (`"#4E4AF6"`) was never exercised. New test asserts a non-`"dark"` theme yields `"#4E4AF6"`.
2. **`Math.abs`.** Every existing fixture used positive formatted amounts, so removing `Math.abs` made no observable difference. New test feeds a negative input amount (raw ratio `-0.5`) and asserts the charted value is the absolute `0.5`.
3. **Single-trade vs merge branch (`length > 1`).** A lone trade at a timestamp must take the verbatim single-trade push, not the weighted-average merge. New test uses a zero output `amount`: the merge branch would divide by it and yield `NaN`, so a finite, unchanged value distinguishes the branches.

## No real bug found

The zero-output-amount case is benign for a single trade (it stays in the verbatim branch), but worth noting: if two trades shared a timestamp *and* their output amounts summed to zero, the weighted-average merge would produce `NaN`. Not exercised in practice / out of scope here — flagging only.

## Inherited-red note

This PR branches off `main`, which is currently red on the 4-fix-set causes (`rs-static`, `sol-static`, `sol-test`, `wasm`) pending #2749-2752. Those reds are inherited by any branch off `main` and are **not** caused by this change. The JS/vitest path for the touched module is green: `npx vitest run src/lib/services/historicalOrderCharts.ts` -> 5 passed. Prettier and eslint are clean on the file. (`svelte-check` reports 2 pre-existing `Cannot find module '@rainlanguage/ui-components'` errors in unrelated `.svelte` files; confirmed identical on the clean baseline — they need the package itself built and are unrelated to this change.)

## Summary by CodeRabbit

* **Tests**
  * Strengthened test coverage for historical order charts functionality. New test cases validate correct chart behavior across different display themes, ensure accurate calculations for trade value ratios regardless of input values, and verify proper handling of edge cases involving trades with zero output amounts.


thedavidmeister:
human-approved; merging

PR #2769, opened 2026-06-16
Add vaultless order IO settings

findolor:
## Summary
- Add `vault-id: vaultless` as the authored settings sentinel for vaultless order IO.
- Reject `vault-id: 0` so vaultless mode must be requested explicitly with the `vaultless` sentinel.
- Represent IO vault settings as `VaultConfig` internally so vaultless and numeric vault IDs are mutually exclusive.
- Update YAML mutation/population helpers so vaultless IO stays vaultless and vault-based IO still auto-populates missing vault IDs.

## Dependent PRs
- Depends on the YAML spec update in [rainlanguage/specs#44](https://github.com/rainlanguage/specs/pull/44).

## Validation
- `cargo check -p raindex_app_settings -p raindex_common`
- `cargo test -p raindex_app_settings vaultless`
- `cargo test -p raindex_app_settings zero_vault_id`
- `cargo test -p raindex_app_settings zero_output_vault_id`
- `cargo test -p raindex_app_settings update_vault_id`
- `cargo test -p raindex_common add_order`

## Review
- Staged local review completed. Initial populate/add-order findings were addressed in stack order; follow-up review was clean.

Refs RAI-1063.

## Summary by CodeRabbit

* **New Features**
  * Orders can now be configured as vaultless, separate from vault-based configurations.

* **Bug Fixes**
  * Zero vault IDs are now rejected during configuration parsing to prevent invalid states.

* **Chores**
  * Updated configuration validation to enforce mutual exclusivity between vault and vaultless settings.

### DIFF
--- a/crates/settings/src/order.rs
+++ b/crates/settings/src/order.rs
@@ -10,7 +10,7 @@ use strict_yaml_rust::{strict_yaml::Hash, StrictYaml};
 use thiserror::Error;
 
 const ALLOWED_ORDER_KEYS: [&str; 5] = ["inputs", "oracle-url", "raindex", "outputs", "rainlang"];
-const ALLOWED_ORDER_IO_KEYS: [&str; 2] = ["token", "vault-id"];
+const ALLOWED_ORDER_IO_KEYS: [&str; 3] = ["token", "vault-id", "vaultless"];
 use wasm_bindgen_utils::{impl_wasm_traits, prelude::*};
 use yaml::{
     context::{Context, OrderBuilderContextTrait, SelectTokensContext},
@@ -32,6 +32,9 @@ impl_wasm_traits!(VaultType);
 pub struct OrderIOCfg {
     pub token_key: String,
     pub token: Option<Arc<TokenCfg>>,
+    #[serde(default)]
+    #[cfg_attr(target_family = "wasm", serde(rename = "vaultless"))]
+    pub vaultless: bool,
     #[cfg_attr(
         target_family = "wasm",
         serde(rename = "vaultId"),
@@ -65,7 +68,77 @@ impl_wasm_traits!(OrderCfg);
 
 impl OrderCfg {
     pub fn validate_vault_id(value: &str) -> Result<U256, ParseOrderConfigSourceError> {
-        U256::from_str(value).map_err(ParseOrderConfigSourceError::VaultParseError)
+        U256::from_str(value)
+            .map_err(ParseOrderConfigSourceError::VaultParseError)
+            .and_then(|vault_id| {
+                if vault_id == U256::ZERO {
+                    Err(ParseOrderConfigSourceError::ZeroVaultId)
+                } else {
+                    Ok(vault_id)
+                }
+            })
+    }
+
+    fn parse_vaultless(value: &StrictYaml, location: &str) -> Result<bool, YamlError> {
+        let value = &value["vaultless"];
+        if value.is_badvalue() {
+            return Ok(false);
+        }
+
+        match value.as_str() {
+            Some("true") => Ok(true),
+            Some(_) | None => Err(YamlError::Field {
+                kind: FieldErrorKind::InvalidValue {
+                    field: "vaultless".to_string(),
+                    reason: "vaultless must be true when present".to_string(),
+                },
+                location: location.to_string(),
+            }),
+        }
+    }
+
+    fn parse_io_vault_fields(
+        value: &StrictYaml,
+        location: &str,
+    ) -> Result<(bool, Option<U256>), YamlError> {
+        let vaultless = Self::parse_vaultless(value, location)?;
+        let vault_id_value = &value["vault-id"];
+        let vault_id = if vault_id_value.is_badvalue() {
+            None
+        } else {
+            Some(
+                vault_id_value
+                    .as_str()
+                    .ok_or_else(|| YamlError::Field {
+                        kind: FieldErrorKind::InvalidType {
+                            field: "vault-id".to_string(),
+                            expected: "a string".to_string(),
+                        },
+                        location: location.to_string(),
+                    })
+                    .and_then(|id| {
+                        Self::validate_vault_id(id).map_err(|e| YamlError::Field {
+                            kind: FieldErrorKind::InvalidValue {
+                                field: "vault-id".to_string(),
+                                reason: e.to_string(),
+                            },
+                            location: location.to_string(),
+                        })
+                    })?,
+            )
+        };
+
+        if vaultless && vault_id.is_some() {
+            return Err(YamlError::Field {
+                kind: FieldErrorKind::InvalidValue {
+                    field: "vaultless".to_string(),
+                    reason: "vaultless cannot be used with vault-id".to_string(),
+                },
+                location: location.to_string(),
+            });
+        }
+
+        Ok((vaultless, vault_id))
     }
 
     pub fn update_vault_id(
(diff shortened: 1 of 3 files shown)


findolor:
* **#2774** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2774?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2773** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2773?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2772** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2772?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2771** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2771?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2770** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2770?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2769** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2769).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios (approved):

crates/settings/src/order.rs
```diff
@@ -32,6 +32,9 @@ impl_wasm_traits!(VaultType);
 pub struct OrderIOCfg {
     pub token_key: String,
     pub token: Option<Arc<TokenCfg>>,
+    #[serde(default)]
```
0xgleb, line 35: should fix: `vaultless: bool` alongside `vault_id: Option<U256>` lets the type hold the illegal `vaultless = true, vault_id = Some(..)` state, which you then rule out at runtime in `parse_io_vault_fields` and guard procedurally in both update helpers. With `#[serde(default)]` a deserialize can also bypass the validated parse path. A small sum type like `enum VaultConfig { Vaultless, WithId(U256), Unset }` makes the illegal combo unrepresentable, deletes the runtime guard, and removes the manual `vaultless: false` you now write at every struct literal.

crates/settings/src/order.rs
```diff
@@ -212,6 +289,121 @@ impl OrderCfg {
...
+            return Err(YamlError::Field {
+                kind: FieldErrorKind::InvalidType {
+                    field: "document".to_string(),
+                    expected: "a map".to_string(),
+                },
+                location: "root".to_string(),
+            });
+        }
+
+        Ok(self.clone())
```
0xgleb, line 404: minor: the body here is a six-level `if let` pyramid where each level's `else` is an error return, which is hard to read. Flattening it with `let ... else` early returns is much cleaner:

```suggestion
        let mut document = self
            .document
            .write()
            .map_err(|_| YamlError::WriteLockError)?;

        let StrictYaml::Hash(ref mut document_hash) = *document else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::InvalidType {
                    field: "document".to_string(),
                    expected: "a map".to_string(),
                },
                location: "root".to_string(),
            });
        };
        let Some(StrictYaml::Hash(ref mut orders)) =
            document_hash.get_mut(&StrictYaml::String("orders".to_string()))
        else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::Missing("orders".to_string()),
                location: "root".to_string(),
            });
        };
        let Some(StrictYaml::Hash(ref mut order)) =
            orders.get_mut(&StrictYaml::String(self.key.to_string()))
        else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::Missing(self.key.clone()),
                location: "orders".to_string(),
            });
        };

        let vec_key = match vault_type {
            VaultType::Input => "inputs",
            VaultType::Output => "outputs",
        };
        let Some(StrictYaml::Array(ref mut vec)) =
            order.get_mut(&StrictYaml::String(vec_key.to_string()))
        else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::Missing(vec_key.to_string()),
                location: format!("order '{0}'", self.key),
            });
        };

        let item_index = vec.iter().position(|item| {
            if let StrictYaml::Hash(ref item_map) = item {
                if let Some(StrictYaml::String(item_token)) =
                    item_map.get(&StrictYaml::String("token".to_string()))
                {
                    return item_token == &token;
                }
            }
            false
        });
        let Some(idx) = item_index else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::InvalidValue {
                    field: vec_key.to_string(),
                    reason: format!("token '{}' not found", token),
                },
                location: format!("order '{0}'", self.key),
            });
        };
        let Some(StrictYaml::Hash(ref mut item_map)) = vec.get_mut(idx) else {
            return Err(YamlError::Field {
                kind: FieldErrorKind::InvalidType {
                    field: vec_key.to_string(),
                    expected: "a hash".to_string(),
                },
                location: format!("order '{0}'", self.key),
            });
        };

        if vaultless {
            item_map.insert(
                StrictYaml::String("vaultless".to_string()),
                StrictYaml::String("true".to_string()),
            );
            item_map.remove(&StrictYaml::String("vault-id".to_string()));
        } else {
            item_map.remove(&StrictYaml::String("vaultless".to_string()));
        }

        let io = match vault_type {
            VaultType::Input => &mut self.inputs[idx],
            VaultType::Output => &mut self.outputs[idx],
        };
        io.vaultless = vaultless;
        if vaultless {
            io.vault_id = None;
        }

        Ok(self.clone())
```

Separately, this entire lock-and-descend skeleton is a verbatim copy of `update_vault_id`; only the leaf mutation differs. A shared private helper that locates the item map and index (e.g. `with_io_item(vault_type, &token, |item_map, idx| ...)`) would let both methods collapse to just their leaf write, though that is a bigger change touching `update_vault_id` too. Worth a compile check after applying the suggestion.

crates/settings/src/order.rs
```diff
@@ -65,7 +68,77 @@ impl_wasm_traits!(OrderCfg);
...
+                                field: "vault-id".to_string(),
+                                reason: e.to_string(),
+                            },
+                            location: location.to_string(),
+                        })
+                    })?,
+            )
+        };
+
+        if vaultless && vault_id.is_some() {
```
0xgleb, line 131: minor: this conflict check runs after `validate_vault_id` has already rejected zero, so `vaultless: true` together with `vault-id: 0` reports `ZeroVaultId` ("use vaultless: true for vaultless IO"), which is confusing advice when the user already set vaultless. Move this check above the vault-id parse, or detect both keys present up front and return the conflict error first.

crates/settings/src/order.rs
```diff
@@ -232,8 +424,13 @@ impl OrderCfg {
                     {
                         for (index, input) in inputs.iter_mut().enumerate() {
                             if let StrictYaml::Hash(ref mut input_hash) = input {
-                                if !input_hash
-                                    .contains_key(&StrictYaml::String("vault-id".to_string()))
+                                if !self
+                                    .inputs
+                                    .get(index)
+                                    .map(|input| input.vaultless)
+                                    .unwrap_or(false)
```
0xgleb, line 431: minor: if the YAML inputs array is ever longer than `self.inputs`, `get(index)` is `None` and `unwrap_or(false)` treats the entry as non-vaultless, injecting a vault-id into what may be a vaultless IO with no error. It can't happen in the normal post-parse flow where lengths match, but an explicit error on mismatch (`.get(index).ok_or_else(...)?.vaultless`) turns a silent wrong result into a loud one. Same pattern applies to the outputs loop just below.

crates/settings/src/order.rs
```diff
@@ -1132,6 +1324,184 @@ orders:
...
+        let order = orders.get("order1").unwrap();
+
+        assert!(order.inputs[0].vaultless);
+        assert_eq!(order.inputs[0].vault_id, None);
+        assert!(!order.outputs[0].vaultless);
+        assert_eq!(order.outputs[0].vault_id, Some(U256::from(2)));
+    }
+
+    #[test]
+    fn test_parse_rejects_zero_vault_id() {
```
0xgleb, line 1368: nit: this rejects `vault-id: 0` only on the input side. `parse_io_vault_fields` runs for outputs too, so a matching case with `vault-id: 0` in `output_extra` (asserting the output index location) would cover the other call site cheaply.

0xgleb (changes requested):

0xgleb (approved):

PR #2778, opened 2026-06-17
fix: mask signer row operand to low byte

### LINKED ISSUE #2653: signers subparser word uses the raw operand as row without & 0xFF masking (unlike signed-context)

thedavidmeister:
## Summary

- `subParserSigners` read the full 16-bit `OperandV2` value as the context row index, so `signer<256>()` silently resolved to the same row as `signer<0>()` (256 mod 256 = 0 at the EVM word level) but only by accident — larger values could land on entirely wrong rows.
- Fix: apply `& 0xFF` to clamp the operand to the intended 0–255 range before passing it to `LibSubParse.subParserContext`.
- Regression test: `testSubParserContextSignerRowMaskLowByte` verifies `signer<256>()` resolves to row 0.
- Pre-pins 0.1.13 deploy constants: SubParser address changes (new bytecode), all other contracts (RaindexV6, RouteProcessor, arb contracts) are unchanged from 0.1.12.

REQUIRES redeploy at land — the SubParser bytecode changes, so it must be redeployed to its new CREATE2 address before `testProdDeploy*` goes green.

Closes https://github.com/rainlanguage/raindex/issues/2653

## Summary by CodeRabbit

* **New Features**
  * Added support for Raindex ecosystem version 0.1.13 with pinned deployment addresses and code hashes for core components.

* **Bug Fixes**
  * Updated operand interpretation in signer-grid row derivation to use low byte only instead of full value.

* **Tests**
  * Added test verifying signer-grid row masking behavior wraps correctly.


thedavidmeister:
🤖 ai:vetter
Reviewed 0b65acef147b447b9337ce9c95545ceb7e965e2a: ready — adds uint256(OperandV2.unwrap(operand)) & 0xFF mask matching sibling subParserSignedContext; test ve
cost 150 — one-line operand mask plus regression test

thedavidmeister:
Design hold (human). The PR premise is incorrect: `LibSubParse.subParserContext` bounds-checks `row > type(uint8).max` and **reverts `ContextGridOverflow` before any truncation**, so on current `main` `signer<256>()` does not "silently resolve to row 0" — it **reverts at parse time**. Applying `& 0xFF` therefore does not make explicit what already happens; it **converts a fail-safe revert into a silent mod-256 wrap** (`signer<256>() → row 0`, `signer<257>() → row 1`, …), letting an out-of-range signers index silently read the wrong signer. The sibling `signed-context` masks to *extract two byte-fields* from a 16-bit operand (never overflows by construction), which does not justify range-clamping `signers` single-field operand from revert to wrap. Parked for a design ruling: keep the fail-safe revert (in which case #2653 is arguably invalid), or accept the silent wrap. Not merged pending that decision.

thedavidmeister:
🤖 ai:producer
Producer note: both red checks (rainix-sol/static + rs-static) are MAIN-STATE, not PR-caused — the no-vm.skip gate flags 5 pre-existing test/lib/deploy files on main (e.g. LibRaindexDeployTaggedConstants.t.sol:24); identical red on main HEAD. rainix-sol/test PASSES here. Greens when main resolves the vm.skip gate.

thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens no source read — the ai:design label's provenance decides this routing, not the diff
Reviewed 37988de2de86bae12ca885d62ffcef97df155dea: needs-work — ROUTING FIX, not a code defect (design doctor, https://github.com/rainlanguage/issue-pr-cron/issues/241): this PR carried ai:design with NO trusted comment raising a design question at its head, which is a state no actor consumes — the human's queue withholds it (there is no claim to present) and every AI actor skips it. It is routed here so it has an owner again. Two moves, both yours: if the PR still needs work, push it — the question was superseded or never recorded, and the push IS the transition. If a genuine design question remains, re-raise it with `pr-review-report flag-design <owner/repo> <pr> "<the question>"`, which puts the row back in the human's queue with a claim behind it. Do not push a no-op commit to clear this.

PR #2786, opened 2026-06-22
fix(webapp): seed selectedChainIds from validChainIds when empty

### LINKED ISSUE #2747: Can't view any orders on the dex without connecting a wallet
On the raindex webapp, the Orders page returns/shows no orders until a wallet is connected — you can't browse or view any orders at all while disconnected.

Orders are public on-chain / subgraph data, so read-only browsing should NOT require a wallet. A wallet should only be needed for actions that move funds or require signing (create order, deposit, withdraw, remove, take/clear).

**Expected:** the Orders list (and order detail pages) load and are fully browsable while wallet-disconnected, across the supported networks. Connecting a wallet additionally surfaces the user's own orders and enables actions — but is not required to see the public order book.

**Actual:** no orders are retrievable/visible until a wallet is connected.

Likely cause to check: the orders query/list is gated behind the connected-account/chain context (e.g. depends on the wallet's chain id or account) rather than the selected network(s); decouple the public read path from wallet state.

thedavidmeister:
## What

Orders page showed nothing for fresh/wallet-disconnected users because `selectedChainIds` started as `[]` and no code ever seeded it.

`validChainIds.subscribe` only filtered stale entries — it never auto-selected chains for users with an empty selection. Connecting a wallet didn't directly fix this either (the wallet sets `chainId`, not `selectedChainIds`). In practice, users only saw orders if they had a prior chain selection persisted in localStorage, which is why the issue correlates with wallet connection: the wallet flow incidentally triggers a chain-selection UI.

## Fix

One-line guard in `validChainIds.subscribe`: when `validChainIds` first becomes non-empty and `selectedChainIds` is empty, seed `selectedChainIds` with all valid chain IDs. Returning users (non-empty `selectedChainIds` in localStorage) keep their stored selection unchanged.

```diff
 validChainIds.subscribe((valid) => {
   if (valid.length > 0) {
     selectedChainIds.update((current) => {
+      if (current.length === 0) return [...valid];
       const filtered = current.filter((id) => valid.includes(id));
       return filtered.length !== current.length ? filtered : current;
     });
   }
 });
```

## Tests

Two new unit tests in `settings.test.ts`:
- `seeds selectedChainIds with all validChainIds when selectedChainIds is empty` — fresh-user path
- `does not re-seed selectedChainIds when validChainIds updates and selectedChainIds is already set` — returning-user path (no clobber)

All seven existing pruning/persistence tests continue to pass.

Closes #2747

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain selection initialization to automatically populate available chains when no explicit selection exists, while preserving user-selected chains during subsequent updates.

### DIFF
--- a/packages/webapp/src/lib/stores/settings.test.ts
+++ b/packages/webapp/src/lib/stores/settings.test.ts
@@ -58,6 +58,26 @@ describe('settings store', () => {
 			validChainIds.set([1, 42161]);
 			expect(get(selectedChainIds)).toEqual([1, 42161]);
 		});
+
+		it('seeds selectedChainIds with all validChainIds when selectedChainIds is empty', () => {
+			// Fresh user: no prior chain selection.
+			// When the client discovers which chains are available, select them all so
+			// orders are visible without requiring a wallet connection first.
+			validChainIds.set([1, 137, 42161]);
+
+			expect(get(selectedChainIds)).toEqual([1, 137, 42161]);
+		});
+
+		it('does not re-seed selectedChainIds when validChainIds updates and selectedChainIds is already set', () => {
+			// Returning user: their explicit selection must not be clobbered when
+			// validChainIds is refreshed (e.g. after a re-init).
+			selectedChainIds.set([1, 42161]);
+
+			validChainIds.set([1, 137, 42161]);
+
+			// 137 must NOT be auto-added; user chose 1 and 42161.
+			expect(get(selectedChainIds)).toEqual([1, 42161]);
+		});
 	});
 
 	describe('selectedChainIds localStorage persistence', () => {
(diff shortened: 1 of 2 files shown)


thedavidmeister:
Reviewed 83f919be835aa61ed36ebea9b29b9f66c1b5cf76: full adversarial review + coverage check. Correct root-cause fix — seeds selectedChainIds from validChainIds when empty so the public order book loads without a wallet (#2747). Two discriminating tests: empty→all-chains seeded; a returning user's explicit selection is NOT clobbered on refresh. Coverage confirmed complete — the Orders LIST consumes selectedChainIds (the bug); the order DETAIL page fetches by URL params, not selectedChainIds, so it already loaded disconnected, and its only wallet use is the remove/deposit/withdraw/take ACTION handlers, exactly what #2747 says should require a wallet. Closes #2747 correct. Minimal (+21/-0), CI green. Note: BEHIND base. Approve. 🤖 ai:assistant

Issue #2799, opened 2026-07-07
Make the deploy-constant check pointer-driven (drop the hardcoded prefix list + FFI/soldeer fetch)


thedavidmeister:
## What

Rework raindex's deploy-constant completeness check to be **pointer-driven** — enumerate the generated `src/generated/*.pointers.sol` and verify each `DEPLOYED_ADDRESS` is pinned in `LibRaindexDeploy.sol` — instead of the current `script/check-published-deploy-constants.sh` (which fetches published soldeer versions via FFI and greps a hardcoded list of six contract prefixes).

## Why

Two brittleness sources the pointer-driven form removes:

1. **Hardcoded contract list.** The six `*_DEPLOYED` prefixes are maintained by hand. Add a seventh deployed contract and the check keeps passing — it never knew to look for it. That's the exact "forgot to pin a constant" gap the check exists to close, reintroduced one level up in the check itself.

2. **Network + FFI dependency.** The check needs FFI and a live registry call, so it `SKIP`s on network flakiness — a guard that can silently not-run. The generated pointer files are already the local, authoritative record of what the current build deploys; comparing them to the lib needs neither.

A pointer-driven check enumerates `src/generated/*.pointers.sol`, extracts each `DEPLOYED_ADDRESS`, and asserts the lib pins that address (case-insensitive on the checksummed hex). It auto-covers every new deployed contract, catches a forgotten *or drifted* address, and runs offline with just `grep` — no FFI, no `SKIP`, and it can drop from a `vm.ffi` test to a plain CI step.

## Scope note (version history)

The current check verifies every *published soldeer version* has its constant suite; pointer-driven verifies the *current build's* pointers are pinned. These converge going forward — bumping the version regenerates the pointers, and the new (unpinned) addresses fail the check until pinned, so "published a new version without constants" is still caught. Frozen historical-version constants are immutable references already verified when they were added, so they need no live re-check. If an explicit per-published-version assertion is still wanted it can sit alongside as a separate concern; the completeness/anti-drift guard is better pointer-driven.

## Context

This is the form being implemented for `st0x.deploy` (which has no FFI). Unifying raindex onto it drops raindex's FFI dependency for this check.

PR #2804, opened 2026-07-12
fix(ci): green main — remove banned vm.skip usages, pin 0.1.13 deploy constants

thedavidmeister:
## What

`main` has been red since 2026-07-07 on `rainix-sol / static / static`, `rs-static / rs-static`, and `rainix-sol / test / test`, and every open PR inherits all three. Two independent root causes, both fixed here:

1. **rainix no-skips static gate** (propagated via the `@main`-pinned reusables): five `vm.skip` occurrences in `test/lib/deploy/` violate "a test either runs and passes, or it is deleted". Fixed per the gate's own policy:
   - `LibRaindexDeployTaggedConstants.t.sol` — dropped the `vm.skip(true)` in the registry-unreachable branch; the `return;` that follows it already produces the identical behavior (nothing to verify, no assertion runs).
   - `LibRaindexDeployIsStartBlockArbitrum.t.sol` — deleted; its single test was `vm.skip(true)` unconditionally (free Arbitrum RPCs cannot serve the `rollFork` historical state), so it has never verified anything. The in-file comment records the start block was verified via manual cast binary search.
   - `LibRaindexDeployStartBlock{Base,Flare,Polygon}.t.sol` — deleted; each was `vm.skip(vm.envOr("CI", false))` (never ran in CI) and, per its own natspec, redundant with the kept `LibRaindexDeployIsStartBlock{Base,Flare,Polygon}` tests that DO run in CI and pin the same `RAINDEX_START_BLOCK_*` constants. `script/build-start-blocks.sh` remains the maintained path for recomputing start blocks after a redeploy. Residual loss: the local-only `forge test` binary-search convenience — the script covers that use.

2. **soldeer tag 0.1.13 published with no pinned deploy constants**: `testAllPublishedSoldeerTagsHaveAFullConstantSuite` (FFI against the live registry) reds main and every PR because PR #2789 (natspec-only edit to `RaindexV6.sol`) published 0.1.13 on 2026-07-03 without the pre-pin the release lifecycle expects. Added the 12 `*_DEPLOYED_{ADDRESS,CODEHASH}_0_1_13` constants to `src/lib/deploy/LibRaindexDeploy.sol`.

## Why no deploy is needed for the 0.1.13 pins

0.1.13 was a natspec-only release: `foundry.toml` has `bytecode_hash = "none"` and `cbor_metadata = false`, so comments cannot change deployed bytecode, and the only main commit since the release (PR #2786) is webapp-only. The 0.1.13 bytecode IS the 0.1.12 deployment, already live on all five chains (`testProdDeploy*` is green on main). The `_0_1_13` values are therefore identical to `_0_1_12`, following the exact precedent of the 0.1.3 / 0.1.8 / 0.1.9 "(Unchanged from …)" pins. Merging this cannot publish a 0.1.14 loop: the autopublish content gate strips pinned `*_<ver>` constant blocks before hashing (rainix#224).

## QA evidence (adversarial-mutation-test pass, scoped to this diff)

- **Baseline:** `forge test --match-contract 'LibRaindexDeployTaggedConstantsTest|LibRaindexDeployTest'` → 32 passed, 0 failed (was: the FFI test failing with the exact `MISSING: …_0_1_13` list on main — CI runs 28851256246 / 29050618038 are the fails-on-base proof for the constants; the static gate's 5-hit `vm.skip` list on those same runs is the fails-on-base proof for part 1).
- **Mutation — constants existence (KILLED):** renaming `RAINDEX_DEPLOYED_ADDRESS_0_1_13` makes `testAllPublishedSoldeerTagsHaveAFullConstantSuite` fail with exactly `MISSING: RAINDEX_DEPLOYED_ADDRESS_0_1_13`; restored, green. The FFI check discriminates per-constant.
- **Mutation — constants VALUE (SURVIVED, known gap):** `script/check-published-deploy-constants.sh` greps only `constant <NAME> =` existence and no test references any versioned pin by name (`grep -rn '_0_1_1[0-3]' test/ script/` → empty), so a corrupted pin value is caught by nothing. This is a pre-existing property of the check for ALL versions, already tracked by #2799 (make the check pointer-driven); not expanded here.
- **Value correctness, established independently of the check:** all six `_0_1_13` (address, codehash) pairs are byte-identical to the floating `DEPLOYED_ADDRESS` / `BYTECODE_HASH` in `src/generated/*.pointers.sol` (verified by direct comparison), and those floating values are validated by the 31 passing `LibRaindexDeployTest` assertions (creation/runtime code hashes recomputed from the committed artifacts). `script/check-published-deploy-constants.sh` → `OK`.
- **No-skips gate:** `grep -rn 'vm.skip' test/ src/` → 0 hits on this branch (5 on main).
- **Adversarial notes:** the `vm.skip`→`return` change preserves the SKIP-path semantics exactly (the only observable difference is forge reporting "pass" instead of "skipped" when the registry is unreachable); the SKIP path itself remains untestable without an FFI mock (pre-existing). Deletions remove zero effective CI coverage (one test never ran anywhere; three never ran in CI and duplicate kept cheap tests).

## REQUIRES redeploy at land

No. This changes no deployed bytecode — constants and test files only.


thedavidmeister:
🤖 ai:vetter
Reviewed 08ece26aaf3d9d03dfbd3392bee72af937054910: ready — green-main fix: 12 0.1.13 pins byte-identical to 0.1.12 and generated pointers (natspec-only release, per documented lifecycle); 5 vm.skip removals verified redundant (kept IsStartBlock* tests pin same constants); QA claims verified in clone; nit: subparser 0.1.13 comment typo 0.1.2->0.1.12
cost 235 — mechanical pins + test deletions, lifecycle context

thedavidmeister:
Superseded by https://github.com/rainlanguage/raindex/pull/2808, which reworks this per review — deploy constants via the rain.factory frozen-snapshot convention (byte-faithful per-tag src/generated/<tag>/ dirs + self-verifying tests, no hand-pinned literals, no registry FFI), and the skipped StartBlock* 'tests' recognised as duplicates of the build-start-blocks.sh script rather than deleted-to-dodge-the-gate. Same two main-red causes fixed, correctly.

PR #2810, opened 2026-07-15
fix: rename the codegen script to script/Build.sol

thedavidmeister:
rainlanguage/rainix#272 (merged, `ef89e90`) made `rainix-copy-artifacts` require the codegen script to be `script/Build.sol`, matched exactly. This repo's script is `script/BuildPointers.sol`, so its `copy-artifacts` job now **errors** until renamed.

That guard replaced an `if: hashFiles('script/BuildPointers.sol')` gate whose failure mode was worse than a red: a repo whose script didn't match that exact name had regeneration **silently skipped**, and the "committed artifacts match freshly built" assert then passed having regenerated nothing — a green over an unchecked tree.

## Change

- `script/BuildPointers.sol` -> `script/Build.sol`, `contract BuildPointers` -> `contract Build`.
- References to the script updated (docs, comments, and any invocation).

The generated `src/generated/*.pointers.sol` files are deliberately **untouched**. Their `AUTOGENERATED BY ./script/BuildPointers.sol` header is emitted by this repo's pinned `rain-sol-codegen`, which still writes that name, so regeneration reproduces them byte-for-byte and `copy-artifacts` stays green. The header and the `.pointers.sol` suffix change when this repo bumps to codegen 0.1.4+ (rainlanguage/rain.sol.codegen#31), which is separate work.

The name: the generated file is a set of constants for a contract, and only the interpreter/word repos generate an actual function pointer table — so "pointers" named the exception rather than the thing.

### DIFF
--- a/.github/workflows/copy-artifacts.yaml
+++ b/.github/workflows/copy-artifacts.yaml
@@ -4,6 +4,6 @@ jobs:
   copy-artifacts:
     # Build from committed sources and assert the committed meta/pointer
     # artifacts are still current. The meta hook is `script/build-meta.sh`;
-    # pointers + abis use the upstream BuildPointers.sol convention.
+    # pointers + abis use the upstream Build.sol convention.
     uses: rainlanguage/rainix/.github/workflows/rainix-copy-artifacts.yaml@main
     secrets: inherit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,8 +41,8 @@ address + codehash. Do the whole cascade without asking:
 
 1. **Regenerate artifacts** — the canonical `rainix-copy-artifacts` sequence, in
    `nix develop github:rainlanguage/rainix#sol-shell`: `forge soldeer install` →
-   `script/build-meta.sh` → `forge script script/Build.sol` →
-   `forge build` → `forge script script/CopyArtifacts.sol --ffi` (if present) →
+   `script/build-meta.sh` → `forge script script/Build.sol` → `forge build` →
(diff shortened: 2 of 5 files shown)


thedavidmeister:
Reviewed 52b07a8ab6f0038176537fd77e49c21fa9684c88: rename only, no behavior change — merge

Diff read in full. Five files, and nothing outside the script's name:

- `script/BuildPointers.sol` -> `script/Build.sol`, `contract BuildPointers` -> `contract Build` (98% similarity, body untouched).
- Four references updated: a comment in `.github/workflows/copy-artifacts.yaml`, a `CLAUDE.md` step (denofmt reflowed the line), a `foundry.toml` comment, and the `RaindexV6FreshTest` doc comment.

No generated artifact, no `src/generated/<tag>/` frozen snapshot dir, and no Solidity outside `script/` is touched — so the rainix#268 append-only gate has nothing to see and no bytecode moves.

Why it is needed: rainlanguage/rainix#272 (merged, `ef89e90`) matches the codegen script by exact name and errors if a repo commits `src/generated/` without `script/Build.sol`. It replaced an `if: hashFiles('script/BuildPointers.sol')` guard whose failure mode was silent — a repo whose script did not match that exact name had regeneration SKIPPED, and the currency check then passed having regenerated nothing.

Checks: 16/16 pass, including `copy-artifacts` with `Regenerate generated sources: success` — the step now runs, so the assert compares a freshly regenerated tree against the committed one. No failing checks; mergeable.

Human approval for merge was given separately ("merge all", over a queue naming this PR). This comment is the hook's required self-review artifact, pinned to the head SHA — not a substitute for that approval.

PR #2825, opened 2026-07-30
perf: add bounded oracle HTTP client

findolor:
## Why

Oracle-backed quote preparation currently creates a fresh HTTP client for every
pair and has no reusable bounded execution primitive. This adds connection-pool
reuse and a conservative per-origin limit before the quote sweep is made
concurrent.

## What changed

- add a reusable `OracleClient` backed by one native/WASM-compatible reqwest
  connection pool
- cap oracle POSTs at eight in flight per origin across independently created clients
- add ordered bounded execution that avoids head-of-line blocking
- route the existing single and batch APIs through the shared client without
  changing their signatures, timeout, response validation, or status handling
- add deterministic tests for the concurrency bound, shared limiter, positional
  stability, partial failure isolation, and slow-leading-request behavior

## Intentionally unchanged

- configured oracle URLs are not assumed to support batch request bodies
- no retries are added for 429/503 or other failures because the endpoint
  contract does not declare POST idempotency
- signed-context and chain quote behavior are unchanged in this prerequisite PR

## Local verification

- `cargo test -p raindex_quote --lib` — 63 passed
- `cargo clippy -p raindex_quote --all-targets --all-features -- -D warnings -D clippy::all`
- `cargo fmt --all -- --check`
- `nix develop .#wasm-shell -c bash -c 'CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test --target wasm32-unknown-unknown --lib -p raindex_quote'`
- `rainix-rs-static`

## Review focus

Please pay particular attention to the per-origin limit of eight, indexed
`buffer_unordered` ordering restoration, and native/WASM client reuse.

## Summary by CodeRabbit

* **Performance**
  * Oracle requests now reuse shared HTTP resources for improved efficiency.
  * Concurrent requests are bounded, with configurable limits for improved stability under load.

* **Reliability**
  * Batch results preserve their original request order.
  * Individual request errors are reported without blocking other requests.
  * Requests now validate endpoints, apply timeouts, and redact sensitive URL details from errors.
  * Response handling and request validation are more consistent.

### DIFF
--- a/crates/quote/src/oracle.rs
+++ b/crates/quote/src/oracle.rs
@@ -1,18 +1,42 @@
 use alloy::primitives::{Address, Bytes, FixedBytes, U256};
 use alloy::sol_types::SolValue;
+use futures::{stream, StreamExt};
+use once_cell::sync::{Lazy, OnceCell};
 use raindex_bindings::IRaindexV6::{OrderV4, SignedContextV1};
 use raindex_subgraph_client::types::common::SgOrder;
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
+use std::{
+    collections::HashMap,
+    future::Future,
+    num::NonZeroUsize,
+    sync::{Arc, Mutex, Weak},
+    time::Duration,
+};
+use tokio::sync::Semaphore;
 use url::Url;
+
+/// Default maximum number of concurrent oracle HTTP requests per caller and
+/// oracle origin.
+///
+/// Eight allows meaningful overlap for quote pages while keeping pressure on
+/// each oracle service conservative. Native clients share each origin's limit
+/// across the process; WASM clients share it within one module instance.
+pub(crate) const ORACLE_REQUEST_CONCURRENCY_LIMIT: usize = 8;
+const ORACLE_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+
+static ORACLE_HTTP_CLIENT: OnceCell<Client> = OnceCell::new();
+type OracleRequestSemaphores = Mutex<HashMap<String, Weak<Semaphore>>>;
+static ORACLE_REQUEST_SEMAPHORES: Lazy<Arc<OracleRequestSemaphores>> =
+    Lazy::new(|| Arc::new(Mutex::new(HashMap::new())));
 
 /// Validate that an oracle URL is safe to POST to.
 /// Only http and https schemes are allowed to prevent SSRF.
-fn validate_oracle_url(url: &str) -> Result<(), OracleError> {
+fn validate_oracle_url(url: &str) -> Result<Url, OracleError> {
     let parsed =
         Url::parse(url).map_err(|e| OracleError::InvalidUrl(format!("Cannot parse URL: {e}")))?;
     match parsed.scheme() {
-        "http" | "https" => Ok(()),
+        "http" | "https" => Ok(parsed),
         scheme => Err(OracleError::InvalidUrl(format!(
             "Unsupported scheme '{scheme}', only http and https are allowed"
         ))),
@@ -22,14 +46,37 @@ fn validate_oracle_url(url: &str) -> Result<(), OracleError> {
 /// Error types for oracle fetching
 #[derive(Debug, thiserror::Error)]
 pub enum OracleError {
-    #[error("HTTP request failed: {0}")]
-    RequestFailed(#[from] reqwest::Error),
+    #[error("HTTP request to {endpoint} failed: {source}")]
+    RequestFailed {
+        endpoint: String,
+        #[source]
+        source: reqwest::Error,
+    },
+
+    #[error("Failed to build oracle HTTP client: {0}")]
+    ClientBuildFailed(#[source] reqwest::Error),
 
     #[error("Invalid oracle URL: {0}")]
     InvalidUrl(String),
 
     #[error("Invalid oracle response: {0}")]
     InvalidResponse(String),
+}
+
+fn request_failed(endpoint: &str, source: reqwest::Error) -> OracleError {
+    OracleError::RequestFailed {
+        endpoint: endpoint.to_owned(),
+        source: source.without_url(),
+    }
+}
+
+fn sanitized_endpoint(url: &Url) -> String {
+    let mut endpoint = url.clone();
+    let _ = endpoint.set_username("");
+    let _ = endpoint.set_password(None);
+    endpoint.set_query(None);
+    endpoint.set_fragment(None);
+    endpoint.into()
 }
 
 /// JSON response format from an oracle endpoint.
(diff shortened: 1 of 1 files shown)


findolor:
* **#2826** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2826?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2825** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2825).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/quote/src/oracle.rs
```diff
@@ -54,6 +75,151 @@ impl From<OracleResponse> for SignedContextV1 {
...
+
+        Ok(response.into_iter().map(Into::into).collect())
+    }
+}
+
+fn build_http_client() -> Result<Client, reqwest::Error> {
+    let builder = Client::builder();
+    #[cfg(not(target_family = "wasm"))]
+    let builder = builder.timeout(std::time::Duration::from_secs(10));
+    builder.build()
```
0xgleb: should fix: the 10s timeout is native-only, but the permit from the shared semaphore is held across the whole `send()` + `error_for_status()` + `json()` body read. On wasm there's no client timeout and no per-request timeout, so a stalled fetch holds its permit forever. Eight of those exhaust the pool and every later oracle request in the module blocks at `acquire().await`, including ones pointing at healthy endpoints. Before this PR a stalled call only affected its own caller, so the shared limiter turns a single hung request into a permanent module-wide stall in the browser build. Either apply `.timeout(Duration::from_secs(10))` per request (reqwest 0.12 backs this with `AbortSignal` on wasm) or wrap the inner future in `tokio::time::timeout` inside `with_request_permit` so the permit is always released.

crates/quote/src/oracle.rs
```diff
@@ -54,6 +75,151 @@ impl From<OracleResponse> for SignedContextV1 {
...
+    /// only its own slot. Requests run through the global concurrency limiter.
+    pub async fn fetch_signed_contexts(
+        &self,
+        requests: Vec<OracleRequest>,
+    ) -> Vec<Result<SignedContextV1, OracleError>> {
+        let requests = requests.into_iter().map(|request| async move {
+            self.fetch_signed_context(&request.url, request.body).await
+        });
+
+        collect_bounded(requests, ORACLE_REQUEST_CONCURRENCY_LIMIT).await
```
0xgleb: should fix: `fetch_signed_contexts` and `OracleRequest` have no callers at this commit and no test. `order_quotes.rs` still awaits `fetch_signed_context` inline inside the nested validInputs x validOutputs loop, so oracle fetches for a quote page are still fully sequential and the headline benefit isn't reachable yet. The untested part is also the interesting part, because `buffer_unordered(8)` and the eight-permit global semaphore are two separate bounds that interact. I'd either wire up the consumer in this PR or keep these `pub(crate)` until it lands, so the published surface only grows once it has a user.

crates/quote/src/oracle.rs
```diff
@@ -32,6 +47,12 @@ pub enum OracleError {
     InvalidResponse(String),
 }
 
+impl From<reqwest::Error> for OracleError {
+    fn from(error: reqwest::Error) -> Self {
+        Self::RequestFailed(error.without_url())
```
0xgleb: should fix: `without_url()` strips the endpoint from every reqwest error, not just the credential. Nothing downstream re-adds it, because `order_quotes.rs` logs only the pair indices and the error, and the URL is per-order metadata. A batch quote spanning orders with different oracles then produces failures reading just `HTTP request failed: HTTP status server error (503)`, with no way to tell which endpoint is down. The redaction goal is right but the granularity is too coarse. You already hold the parsed `Url` in `fetch_responses`, so build a sanitized endpoint (clear query, username, password) and make the variant `RequestFailed { endpoint, #[source] source }` constructed there rather than via a blanket `From`. Worth extending the redaction test to assert host and path survive while the api_key doesn't.

crates/quote/src/oracle.rs
```diff
@@ -1,18 +1,33 @@
...
 
+/// Maximum number of oracle HTTP requests in flight across the process.
+///
+/// Eight allows meaningful overlap for quote pages while keeping pressure on
+/// shared oracle services conservative. The same limit applies across all
+/// [`OracleClient`] instances.
+pub const ORACLE_REQUEST_CONCURRENCY_LIMIT: usize = 8;
+
+static ORACLE_HTTP_CLIENT: OnceCell<Client> = OnceCell::new();
+static ORACLE_REQUEST_SEMAPHORE: Lazy<Arc<Semaphore>> =
```
0xgleb: should fix: this semaphore is keyed on nothing, so eight in-flight requests to one unresponsive oracle block requests to every other oracle in the process. Orders carry their own URL via `extract_oracle_url`, so unrelated healthy endpoints queue behind up to 10s of dead time per slot on native. That reintroduces across endpoints exactly the head-of-line blocking `collect_bounded` removes within a batch. Keying the limiter by `url.origin().ascii_serialization()` in a map of semaphores, acquired in `fetch_responses` after `validate_oracle_url` hands back the parsed `Url`, would keep the cap per host instead.

crates/quote/src/oracle.rs
```diff
@@ -54,6 +75,151 @@ impl From<OracleResponse> for SignedContextV1 {
...
+/// concurrency limiter.
+#[derive(Clone)]
+pub struct OracleClient {
+    client: Client,
+    request_semaphore: Arc<Semaphore>,
+}
+
+/// One independent oracle POST prepared for bounded execution.
+pub struct OracleRequest {
+    url: String,
```
0xgleb: minor: `OracleRequest` holds `url: String` even though `validate_oracle_url` now returns a parsed `Url`, so that value gets thrown away and re-derived on every fetch. `OracleRequest::new("not-a-url".into(), body)` constructs fine and only fails once the batch is being driven, surfacing as one error slot among many rather than at the point the caller made the mistake. Storing the validated `Url` and validating in the constructor would move that failure to where it can be acted on.

crates/quote/src/oracle.rs
```diff
@@ -1,18 +1,33 @@
...
+use std::{future::Future, sync::Arc};
+use tokio::sync::Semaphore;
 use url::Url;
 
+/// Maximum number of oracle HTTP requests in flight across the process.
+///
+/// Eight allows meaningful overlap for quote pages while keeping pressure on
+/// shared oracle services conservative. The same limit applies across all
+/// [`OracleClient`] instances.
+pub const ORACLE_REQUEST_CONCURRENCY_LIMIT: usize = 8;
```
0xgleb: minor: there's no way to raise or lower this per caller, which is inconsistent with how the same quote sweep already handles concurrency, since `get_order_quotes` threads a caller-supplied `chunk_size: Option<usize>` for RPC batching. Once the sweep goes concurrent, a bulk CLI run and an interactive webapp request share one fixed budget of 8 with no fairness, and a workload needing a different bound has to patch and release this crate. Threading a limit through `OracleClient::new` defaulting to 8 would match the existing pattern.

crates/quote/src/oracle.rs
```diff
@@ -54,6 +75,151 @@ impl From<OracleResponse> for SignedContextV1 {
...
+where
+    F: Future<Output = T>,
+{
+    let mut results = stream::iter(
+        futures
+            .into_iter()
+            .enumerate()
+            .map(|(index, future)| async move { (index, future.await) }),
+    )
+    .buffer_unordered(limit.max(1))
```
0xgleb: minor: `limit: usize` admits `0`, and `.max(1)` silently rewrites it to 1 instead of rejecting it. The guard exists because the type allows a value the function can't honour. `NonZeroUsize` makes it unrepresentable, drops the fudge, and puts the contract in the signature.

crates/quote/src/oracle.rs
```diff
@@ -1,18 +1,33 @@
...
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
+use std::{future::Future, sync::Arc};
+use tokio::sync::Semaphore;
 use url::Url;
 
+/// Maximum number of oracle HTTP requests in flight across the process.
+///
+/// Eight allows meaningful overlap for quote pages while keeping pressure on
+/// shared oracle services conservative. The same limit applies across all
```
0xgleb: nit: on wasm each module instantiation gets its own copy of these statics, so the ceiling is per instance rather than per process or per tab. Since the webapp is a first-class consumer here, someone sizing oracle capacity from this comment would underestimate peak concurrency by the number of live instances. Worth saying per module instance.

0xgleb (changes requested):

JuaniRios (approved):

crates/quote/src/oracle.rs
```diff
@@ -54,6 +101,211 @@ impl From<OracleResponse> for SignedContextV1 {
...
+            })
+    }
+
+    async fn with_request_permit<F, T>(&self, url: &Url, request: F) -> T
+    where
+        F: Future<Output = T>,
+    {
+        let request_semaphore = self.request_semaphore(url);
+        let _permit = request_semaphore
+            .acquire()
```
0xgleb, line 182: minor: acquire() waits outside the 10s reqwest timeout, so under same-origin contention a call can sit in the semaphore queue for much longer than ORACLE_REQUEST_TIMEOUT before send() even starts. Cancellation still works if the parent drops, and other origins stay isolated, but the constant is not an end-to-end bound. Worth documenting that, or giving acquire its own deadline once the concurrent sweep lands.

0xgleb (approved):

PR #2826, opened 2026-07-30
perf: prepare oracle quote contexts concurrently

findolor:
## Why

Quote-target construction currently awaits each oracle POST inside the nested order/input/output loops. Production evidence from the st0x REST API showed about 49.6 oracle requests and about 9.7 seconds of target-build latency per cache miss, while the subsequent batched chain quote usually took only 100–600 ms.

## What changed

- prepare every valid pair and its established single-pair oracle request body in positional order
- execute oracle POSTs through the shared process-wide client with a conservative global concurrency limit of 8
- restore results to their original request slots before composing quote targets
- preserve oracle-before-injector signed-context order
- isolate an oracle error to only its corresponding pair response
- keep the chain quote as one `BatchQuoteTarget::do_quote` operation
- add tracing for oracle request, endpoint, success/failure, concurrency-limit, batching, and duration data
- add a deterministic end-to-end test that asserts exact oracle request bodies, partial failure isolation, exact ordered RPC multicall contents, signed-context composition, distinct RPC-result slot mapping, and a single chain RPC request

## Deliberate constraints

- The existing batch encoder/API is not used automatically because order metadata exposes only an oracle URL and does not declare that the configured endpoint accepts the batch ABI body.
- Requests remain individual POSTs and are bounded globally at 8; there is no unbounded fan-out.
- No retry/backoff was added because the endpoint contract does not declare POST idempotency. Existing 429/503 status handling remains a per-pair failure.
- Public response contracts, native/WASM support, timeout behavior, and chain RPC batching semantics are unchanged.

## Verification

- `cargo test -p raindex_quote --lib`
- `cargo clippy -p raindex_quote --all-targets --all-features -- -D warnings -D clippy::all`
- `cargo fmt --all -- --check`
- `nix develop .#wasm-shell -c bash -c 'CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test --target wasm32-unknown-unknown --lib -p raindex_quote'`
- `nix develop .#wasm-shell -c rainix-rs-static`
- `git diff --check`

## Stack

Depends on #2825, which adds the reusable shared HTTP client and bounded ordered oracle-request primitive.

### DIFF
--- a/crates/quote/src/order_quotes.rs
+++ b/crates/quote/src/order_quotes.rs
@@ -3,6 +3,7 @@ use crate::injector::NoopInjector;
 use crate::{
     error::Error,
     injector::SignedContextInjector,
+    oracle::{OracleBatchRequest, OracleClient, ORACLE_REQUEST_CONCURRENCY_LIMIT},
     quote::{BatchQuoteTarget, QuoteTarget},
     OrderQuoteValue,
 };
@@ -12,7 +13,7 @@ use raindex_bindings::provider::mk_read_provider;
 use raindex_bindings::IRaindexV6::{OrderV4, QuoteV2, SignedContextV1};
 use raindex_subgraph_client::types::common::SgOrder;
 use serde::{Deserialize, Serialize};
-use std::str::FromStr;
+use std::{collections::HashMap, str::FromStr};
 use tracing::{debug, info, warn};
 #[cfg(target_family = "wasm")]
 use wasm_bindgen_utils::prelude::js_sys::Date;
@@ -54,6 +55,26 @@ impl QuoteTiming {
     }
 }
 
+struct QuotePairPreparation {
+    pair: Pair,
+    quote_target: QuoteTarget,
+    oracle_url: Option<String>,
+}
+
+struct OracleBatchItem {
+    pair_preparation_index: usize,
+}
+
+struct OracleBatchPreparation {
+    url: String,
+    items: Vec<OracleBatchItem>,
+}
+
+struct QuotedPairMetadata {
+    slot_index: usize,
+    pair: Pair,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(target_family = "wasm", derive(Tsify))]
 #[serde(rename_all = "camelCase")]
@@ -92,12 +113,13 @@ impl_wasm_traits!(Pair);
 /// meta and appending any caller-supplied injector contexts.
 ///
 /// For each order, if the meta contains a `RaindexSignedContextOracleV1`
-/// entry, the oracle URL is extracted and signed context is fetched per IO
-/// pair via POST. Any additional entries produced by `injector` are appended
-/// after the oracle entries (composition order: `[oracle..., injected...]`),
-/// and the composed list is attached to the `QuoteV2.signedContext` before
-/// the multicall is issued. This matters for gated orders whose
-/// `calculate-io` asserts on signed context during quoting.
+/// entry, the oracle URL is extracted and every IO pair for the same exact URL
+/// is fetched in one batch POST. Different URL batches execute concurrently.
+/// Any additional entries produced by `injector` are appended after the oracle
+/// entries (composition order: `[oracle..., injected...]`), and the composed
+/// list is attached to the `QuoteV2.signedContext` before the multicall is
+/// issued. This matters for gated orders whose `calculate-io` asserts on
+/// signed context during quoting.
 pub async fn get_order_quotes(
     orders: Vec<SgOrder>,
     block_number: Option<u64>,
@@ -139,22 +161,15 @@ pub async fn get_order_quotes(
         "resolved quote block number"
     );
 
-    // Responses are assembled in strict iteration order. Pairs whose oracle
-    // fetch failed get a failure response stored immediately; quoted pairs
-    // leave a hole that is filled in after the RPC batch returns. This
-    // preserves per-order positional alignment for callers that re-slice the
-    // flat response vector by per-order pair counts.
-    let mut all_responses: Vec<Option<BatchOrderQuotesResponse>> = Vec::new();
-    // Parallel tracking for the subset of iteration slots that were quoted,
-    // so we can scatter RPC results back into `all_responses`.
-    let mut all_pairs: Vec<Pair> = Vec::new();
-    let mut all_quote_targets: Vec<QuoteTarget> = Vec::new();
-    let mut all_signed_contexts: Vec<Vec<SignedContextV1>> = Vec::new();
-    let mut quoted_slot_indices: Vec<usize> = Vec::new();
+    // Pair metadata is prepared in strict iteration order. Oracle pairs are
+    // grouped by exact endpoint URL so each endpoint receives one batch POST.
+    // Different endpoint batches execute concurrently below, then their
+    // results are scattered back into pair order before the batched chain
+    // quote.
+    let mut pair_preparations: Vec<QuotePairPreparation> = Vec::new();
+    let mut oracle_batches: Vec<OracleBatchPreparation> = Vec::new();
+    let mut oracle_batch_indices: HashMap<String, usize> = HashMap::new();
     let mut skipped_self_trade_pair_count = 0usize;
-    let mut oracle_fetch_count = 0usize;
-    let mut oracle_fetch_failure_count = 0usize;
-    let mut injected_context_count = 0usize;
 
     let target_build_started_at = QuoteTiming::now();
     for order in &orders {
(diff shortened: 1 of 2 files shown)


findolor:
* **#2826** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2826?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2826?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2825** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2825?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2826).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
+
+        match oracle_context {
+            Ok(oracle_context) => {
+                if preparation.oracle_url.is_some() {
+                    debug!(
+                        raindex = %raindex,
+                        input_index,
+                        output_index,
+                        "fetched quote oracle context"
+                    );
```
0xgleb, line 368: minor: the old code timed each oracle POST and put `duration_ms` on both the success debug and the failure warn. Those are gone, and neither the new debug nor the warn carries a duration or an endpoint, so the only timing left is the aggregate on the completed-fetches span, which is dominated by the slowest request in the batch. Given the change is motivated by production evidence of 49.6 requests and 9.7s per cache miss, losing per-request attribution removes exactly the signal that diagnosed the problem. I'd have `fetch_signed_contexts` return elapsed time per request, or emit a debug with host and duration inside the per-request future, and put `duration_ms` plus the endpoint host back on the per-pair lines.

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
+    let oracle_fetch_count = oracle_requests.len();
+    let oracle_fetch_started_at = QuoteTiming::now();
+    info!(
+        oracle_request_count = oracle_fetch_count,
+        oracle_endpoint_count = oracle_urls.len(),
+        oracle_concurrency_limit = ORACLE_REQUEST_CONCURRENCY_LIMIT,
+        oracle_batch_request_count = 0,
+        "starting bounded quote oracle context fetches"
+    );
+    let oracle_results: Vec<Result<SignedContextV1, String>> = if oracle_requests.is_empty() {
```
0xgleb: minor: before this change each `get_order_quotes` call awaited oracle POSTs serially, so it held at most one global permit at a time. Now one call runs its whole pair set through `buffer_unordered(8)` against the same eight-permit process-wide semaphore, so a call whose order meta points at a dead endpoint fills every permit with requests that each sit for the full 10s timeout, and refills them wave after wave at roughly 50 requests per page. Every other concurrent quote request in the process blocks on `acquire()`, which has no deadline and isn't covered by the reqwest timeout because the timeout only starts once the permit is held. Order meta is caller-supplied, so on the multi-tenant REST API this is one order degrading everyone. Capping in-flight requests per URL as well as globally, and putting a deadline around the permit acquisition so a stalled endpoint surfaces as a per-pair failure, would contain it.

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
+
+    for preparation in pair_preparations {
+        let oracle_context = if preparation.oracle_url.is_some() {
+            oracle_results
+                .next()
+                .unwrap_or_else(|| Err("Missing oracle response slot".to_string()))
+                .map(|context| vec![context])
+        } else {
+            Ok(vec![])
+        };
```
0xgleb: minor: the mapping from oracle responses back to pairs is positional by convention. The build loop pushes under `if let Some(url) = &oracle_url` and a separate later loop drains `oracle_results.next()` under `preparation.oracle_url.is_some()`, with nothing tying them together except both re-deriving the same predicate. It's correct today, but the `unwrap_or_else(|| Err("Missing oracle response slot"))` fallback only fires for trailing pairs after the iterator empties, so if a later change ever skips or dedupes a request in the build loop, every subsequent pair silently gets the previous pair's signed context and only the last one errors. That context goes on-chain as `QuoteV2.signedContext`, so a shift would attach an attestation to an order it was never issued for. Storing an `oracle_request_index: Option<usize>` on the preparation and indexing by it, plus a hard error on a length mismatch before the loop, makes the binding explicit and fails loudly.

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
-                            signed_context: vec![],
-                        }));
+                // Oracle context remains live even when the quote is pinned to
+                // a historical block. Configured endpoints are sent only the
+                // established single-pair body because their metadata does not
+                // declare batch-body compatibility.
+                if let Some(url) = &oracle_url {
+                    if !oracle_endpoint_counted {
+                        oracle_urls.insert(url.clone());
+                        oracle_endpoint_counted = true;
```
0xgleb: minor: `HashSet::insert` is already idempotent, so this flag only saves a `String` clone, and `oracle_url` is loop-invariant across both inner loops anyway. Moving the insert up next to the extraction drops the flag entirely and keeps the inner loop about per-pair work. Worth noting it would then also count an endpoint for an order whose pairs are all self-trades and skipped, so if that distinction matters, deriving the count at log time from the prepared pairs avoids both the flag and the clones.

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
+        }
+    }
+
+    let oracle_fetch_count = oracle_requests.len();
+    let oracle_fetch_started_at = QuoteTiming::now();
+    info!(
+        oracle_request_count = oracle_fetch_count,
+        oracle_endpoint_count = oracle_urls.len(),
+        oracle_concurrency_limit = ORACLE_REQUEST_CONCURRENCY_LIMIT,
+        oracle_batch_request_count = 0,
```
0xgleb: minor: this is a literal, and per the deliberate constraints the batch encoder is never invoked from this path, so the field can't be anything but 0 while the code is in this shape. The description lists batching as one of the tracing goals, so someone building a dashboard on it would be measuring nothing and might conclude batching is broken. I'd drop the field until batch-endpoint support exists, or compute it from pairs whose metadata declares batch compatibility so at least the intent is visible in the expression.

crates/quote/src/order_quotes.rs
```diff
@@ -195,100 +201,161 @@ pub async fn get_order_quotes(
...
+                .fetch_signed_contexts(oracle_requests)
+                .await
+                .into_iter()
+                .map(|result| result.map_err(|error| error.to_string()))
+                .collect(),
+            Err(error) => {
+                let error = error.to_string();
+                (0..oracle_fetch_count)
+                    .map(|_| Err(error.clone()))
+                    .collect()
```
0xgleb: nit: the range index is discarded here, so this is `vec!`'s repeat form written as an iterator chain, and it clones the error once per element including the last. `vec![Err(error); oracle_fetch_count]` says the same thing and clones n-1 times.

crates/quote/src/order_quotes.rs
```diff
@@ -406,6 +485,157 @@ mod tests {
...
+                let (mut stream, _) = listener.accept().await.unwrap();
+                let routes = routes.clone();
+                handlers.push(tokio::spawn(async move {
+                    let body = read_http_body(&mut stream).await;
+                    let response = routes
+                        .iter()
+                        .find_map(|(expected_body, response)| {
+                            (expected_body == &body).then_some(response)
+                        })
+                        .and_then(Option::as_ref);
```
0xgleb: nit: `then_some` yields `Option<&Option<OracleResponse>>` and the trailing `and_then(Option::as_ref)` exists only to flatten it, which collapses two different meanings of none: no route matched the body, and a route matched but is configured to fail. The test asserts on a 503, so that distinction is load-bearing. If `encode_oracle_body` ever changes encoding, an unmatched body falls through to the same 503 path and the test keeps passing for the wrong reason. Using `find` and borrowing the payload drops the nested Option, and naming the two outcomes (respond vs unavailable) with a panic on no match would make an encoding drift fail loudly.

0xgleb (approved):

JuaniRios (approved):

PR #2827, opened 2026-07-31
Fix orders token filter source parity

findolor:
## Why

`GetOrdersFilters` currently has source-dependent semantics when both token lists are non-empty and represent the same set. The local SQLite query canonicalizes equal input/output sets and matches the token on either side, while the subgraph query puts `inputs_` and `outputs_` on one `Order_filter`, which GraphQL combines with AND.

A live Base (chain 8453) reproduction against the production registry and a fresh SQLite database isolated the mismatch:

- Directional USDC filters matched exactly across sources: active input-only 45/45, active output-only 48/48, all input-only 532/532, and all output-only 572/572, including order hashes.
- Equal USDC lists diverged: active subgraph/local DB 1/92 and all 7/1,097.
- Equal USDC+WETH lists also diverged: active 6/92 and all 74/1,098.

This blocks ST0x-Technology/st0x.rest.api#166, where omitted `side` intentionally maps to the same token set for inputs and outputs.

## What changed

- Exposed the pinned schema's recursive `or: [Order_filter]` field in `SgOrdersListQueryFilters`.
- Canonicalized non-empty input/output token sets by sorting and deduplicating before equality comparison.
- Emitted equal sets with `or` as the only top-level field, distributing owners, active, order hash, raindex, and positive output-vault balance into both branches to satisfy Goldsky's restriction against mixing column filters with `or`.
- Combined output-token membership and positive-balance fields in the output branch's single `outputs_` filter.
- Preserved one-sided filters and unequal two-sided directional AND filters.
- Added focused subgraph tests for reordered/duplicated equal sets, one-sided and unequal sets, plus a serialized request-level assertion covering the Goldsky-compatible OR shape with representative common filters and positive output balance.
- Strengthened the local DB test to lock the same reordered/duplicated canonical-set trigger.

## Intentionally unchanged

- Native local-DB readiness, query-source selection, and fallback policy.
- Pagination and unrelated SDK behavior.
- Downstream REST code and the st0x REST submodule pointer.
- No multi-query or N+1 workaround.

ST0x REST will update its submodule only after Arda evaluates and merges this upstream PR.

## Verification

- `cargo test -p raindex_subgraph_client`
- `cargo test -p raindex_common canonical_equal_input_output_tokens_use_single_combined_exists`
- `cargo test --workspace`
- `nix develop .#wasm-shell -c rainix-rs-static`
- Exact `wasm-test` workflow command for `raindex_quote`, `raindex_bindings`, `raindex_js_api`, and `raindex_common`
- Exact release-WASM artifact build for `raindex_js_api`
- Exact JS bindings workspace install/build/test workflow: 6 files and 124 tests passed
- `git diff --check main...HEAD`

## Reviewer focus

Please focus on the distributed OR composition: nothing except `or` may remain at the top level, every common filter must appear in both branches, and the output branch must combine token membership with positive output balance in one `outputs_` filter.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved order filtering by consistently handling token lists regardless of ordering or duplicate entries.
  - Equal input and output token selections now match orders in either direction.
  - Preserved directional matching when token selections differ or only one side is specified.
  - Combined token filters now work correctly with owner, activity, hash, source, and balance filters.
  - Active-order filtering now returns only active orders.

### DIFF
--- a/crates/subgraph/src/raindex_client/order.rs
+++ b/crates/subgraph/src/raindex_client/order.rs
@@ -1,11 +1,102 @@
 use super::*;
-use alloy::{
-    hex,
-    primitives::{B256, U256},
-};
-use rain_math_float::Float;
+use crate::utils::float::F0;
+use alloy::{hex, primitives::B256};
+use std::collections::BTreeSet;
 
 impl RaindexSubgraphClient {
+    fn orders_list_filter(filter_args: SgOrdersListFilterArgs) -> Option<SgOrdersListQueryFilters> {
+        let tokens = filter_args.tokens.as_ref();
+        let input_tokens = tokens.map_or(&[][..], |tokens| tokens.inputs.as_slice());
+        let output_tokens = tokens.map_or(&[][..], |tokens| tokens.outputs.as_slice());
+        let has_input_tokens = !input_tokens.is_empty();
+        let has_output_tokens = !output_tokens.is_empty();
+        let has_positive_output_vault_balance =
+            filter_args.has_positive_output_vault_balance == Some(true);
+        let has_filters = !filter_args.owners.is_empty()
+            || filter_args.active.is_some()
+            || filter_args.order_hash.is_some()
+            || !filter_args.raindexes.is_empty()
+            || has_input_tokens
+            || has_output_tokens
+            || has_positive_output_vault_balance;
+
+        has_filters.then(|| {
+            let canonicalize = |tokens: &[String]| {
+                tokens
+                    .iter()
+                    .cloned()
+                    .collect::<BTreeSet<_>>()
+                    .into_iter()
+                    .collect::<Vec<_>>()
+            };
+            let equal_tokens = if has_input_tokens && has_output_tokens {
+                if input_tokens == output_tokens {
+                    Some(canonicalize(input_tokens))
+                } else {
+                    let inputs = canonicalize(input_tokens);
+                    (inputs == canonicalize(output_tokens)).then_some(inputs)
+                }
+            } else {
+                None
+            };
+            let positive_vault_id = has_positive_output_vault_balance
+                .then(|| SgBytes(hex::encode_prefixed(B256::ZERO)));
+            let positive_balance = has_positive_output_vault_balance.then(|| SgBytes(F0.as_hex()));
+            let vault_filter = |token_in| SgVaultTokenFilter {
+                token_in,
+                vault_id_not: positive_vault_id.clone(),
+                balance_gt: positive_balance.clone(),
+            };
+            let common_filter = SgOrdersListQueryFilters {
+                owner_in: filter_args.owners,
+                active: filter_args.active,
+                order_hash: filter_args.order_hash,
+                outputs_: has_positive_output_vault_balance.then(|| vault_filter(vec![])),
+                raindex_in: filter_args.raindexes,
+                ..Default::default()
+            };
+
+            match equal_tokens {
+                None => {
+                    let directional_filter = SgOrdersListQueryFilters {
+                        inputs_: has_input_tokens.then(|| SgVaultTokenFilter {
+                            token_in: input_tokens.to_vec(),
+                            vault_id_not: None,
+                            balance_gt: None,
+                        }),
+                        ..common_filter
+                    };
+
+                    if has_output_tokens {
+                        SgOrdersListQueryFilters {
+                            outputs_: Some(vault_filter(output_tokens.to_vec())),
+                            ..directional_filter
+                        }
+                    } else {
+                        directional_filter
+                    }
+                }
+                Some(tokens) => SgOrdersListQueryFilters {
+                    or: Some(vec![
+                        SgOrdersListQueryFilters {
+                            inputs_: Some(SgVaultTokenFilter {
+                                token_in: tokens.clone(),
+                                vault_id_not: None,
+                                balance_gt: None,
+                            }),
+                            ..common_filter.clone()
+                        },
+                        SgOrdersListQueryFilters {
+                            outputs_: Some(vault_filter(tokens)),
+                            ..common_filter
+                        },
+                    ]),
+                    ..Default::default()
+                },
+            }
+        })
+    }
+
     /// Fetch single order
     pub async fn order_detail(&self, id: &Id) -> Result<SgOrder, RaindexSubgraphClientError> {
         let data = self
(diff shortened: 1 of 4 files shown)


findolor:
* **#2827** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2827?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2827).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/subgraph/src/raindex_client/order.rs
```diff
@@ -4,8 +4,97 @@ use alloy::{
...
+                            }),
+                            ..common_filter.clone()
+                        },
+                        SgOrdersListQueryFilters {
+                            outputs_: Some(vault_filter(tokens)),
+                            ..common_filter
+                        },
+                    ]),
+                    ..Default::default()
+                },
```
JuaniRios, line 95: should fix: the schema pins that `Order_filter.or` exists, but nothing pins how graph-node evaluates `or` when each branch holds a child-entity filter. If the lowering fans out and a single order matches both branches, one order can occupy two rows in a page, which duplicates it and drops the last row of the page. All new tests are hand-built: the httpmock test returns an empty canned response for any body, so it cannot catch this. `vaults_list` runs the same shape in production, which is good evidence but not proof. Record one real Goldsky response for this exact body, with an order that holds the same token on both sides, and assert the page content. That retires the question for both call sites.

crates/subgraph/src/raindex_client/order.rs
```diff
@@ -187,6 +228,154 @@ mod tests {
...
+            then.status(200).json_body(json!({"data": {"orders": []}}));
+        });
+
+        let result = client
+            .orders_list(
+                SgOrdersListFilterArgs {
+                    owners: vec![SgBytes(owner.to_string())],
+                    active: Some(true),
+                    order_hash: Some(SgBytes(order_hash.to_string())),
+                    tokens: Some(SgOrdersTokensFilterArgs {
```
JuaniRios: minor: this test is valuable, but all of its checks live in one boolean `.matches()` closure. When one condition fails, the test only says the mock did not match, with no field name and no expected-versus-actual values. Capture the request body outside the matcher and assert each field with `assert_eq!` so a failure names the broken field.

crates/subgraph/src/raindex_client/order.rs
```diff
@@ -4,8 +4,97 @@ use alloy::{
...
+                    inputs_: has_input_tokens.then(|| SgVaultTokenFilter {
+                        token_in: input_tokens.to_vec(),
+                        vault_id_not: None,
+                        balance_gt: None,
+                    }),
+                    outputs_: if has_output_tokens {
+                        Some(vault_filter(output_tokens.to_vec()))
+                    } else {
+                        common_filter.outputs_.clone()
+                    },
```
JuaniRios: nit: the `else` arm clones `common_filter.outputs_` only to re-assign the value that `..common_filter` already supplies. Omit the field and let struct-update syntax provide it; only set `outputs_` when there is an override.

crates/subgraph/src/raindex_client/order.rs
```diff
@@ -4,8 +4,97 @@ use alloy::{
...
+            let equal_tokens = if has_input_tokens && has_output_tokens {
+                if input_tokens == output_tokens {
+                    Some(canonicalize(input_tokens))
+                } else {
+                    let inputs = canonicalize(input_tokens);
+                    (inputs == canonicalize(output_tokens)).then_some(inputs)
+                }
+            } else {
+                None
+            };
```
JuaniRios, line 41: minor: the filter meaning now flips implicitly on set equality, in two independent implementations (here and `fetch_orders_common.rs`). `{inputs: [USDC], outputs: [USDC]}` means OR; add one token to one side and the whole filter silently becomes AND, which can collapse the result set. There is also no way left to express a genuine both-sides AND for equal sets. A follow-up is fine: add an explicit `TokenMatch::EitherSide | Directional` selector, default it to `EitherSide` on equal sets to keep this behavior, and branch on it in both implementations.

crates/subgraph/src/raindex_client/order.rs
findolor: Thanks — I checked both the lowering and the deployed behavior. graph-node lowers each child-entity filter to a correlated `EXISTS` and combines the branches with SQL `OR`, so an order matching both branches cannot fan out the root result rows. I also ran this exact distributed-OR shape against the pinned production Base Goldsky endpoint: the first 1,000 results had 1,000 unique IDs, including seven orders that matched USDC on both input and output. Pinning one of those order hashes in both branches returned exactly one row. A recorded response fixture would only assert canned JSON rather than exercise Goldsky semantics, so I’m not adding one here.

crates/subgraph/src/raindex_client/order.rs
findolor: Agreed that separate field assertions would produce better failure diagnostics. The current request-level test still covers the exact serialized filter shape, and this is test ergonomics rather than a correctness gap, so I’m keeping this focused parity PR unchanged and treating the capture/assertion refactor as follow-up cleanup.

crates/subgraph/src/raindex_client/order.rs
findolor: The clone is theoretically avoidable, but `outputs_` has to be conditionally overridden when output tokens are present. Simply omitting the field only works for the no-output-token case; removing the clone requires an extra branch or intermediate struct. I’m keeping the current expression because it is more direct and the `Option<SgVaultTokenFilter>` clone is negligible here.

crates/subgraph/src/raindex_client/order.rs
findolor: Agreed that an explicit selector could improve expressiveness. This PR intentionally restores the semantics already implemented by the local DB and required for source parity: equal non-empty sets mean either side, while unequal two-sided sets remain directional AND. `TokenMatch` would be a separate public API expansion across Rust, WASM/TypeScript, and downstream callers, so I’m deferring that to a follow-up product/API decision rather than expanding this fix.

crates/subgraph/src/raindex_client/order.rs
findolor: Addressed in ea58f4881. The directional path now constructs the input/common filter first and conditionally overlays `outputs_`, avoiding the clone while preserving the existing positive-output-balance filter when no output token set is supplied.

crates/subgraph/src/raindex_client/order.rs
findolor: Addressed in ea58f4881. The request-level test now builds and serializes the exact Cynic operation directly, then uses separate assertions for the top-level `or`-only shape, both distributed common-filter branches, and each nested token/balance constraint. This keeps the production-shape coverage while making failures diagnostic.

JuaniRios (approved):

PR #2831, opened 2026-08-13
chore: bump soldeer rain-math-binary 0.1.1 -> 0.1.3

thedavidmeister:
Bumps the Soldeer dependency `rain-math-binary` from `0.1.1` to `0.1.3`.

The upstream `0.1.1 -> 0.1.3` source delta is NatSpec/comment-only in
`src/lib/LibCtPop.sol` — no behavioural change, so no bytecode, pointer or gas
movement is expected from the library itself.

This repo uses version-suffixed Soldeer remappings, so the version string moves
in `foundry.toml`, `remappings.txt` and `soldeer.lock` together. That form is
preserved deliberately:
`rain-math-binary-0.1.3/=dependencies/rain-math-binary-0.1.3/`. No unversioned
alias and no `0.1.1 -> 0.1.3` compatibility shim was introduced.

No import statement in this repo's own `src/`, `test/` or `script/` references
`rain-math-binary` at all (`git grep LibCtPop -- src test script` is empty), so
there are no import rewrites in this diff.

## Diff

Three files, nothing else:

- `foundry.toml` — `rain-math-binary = "0.1.1"` -> `"0.1.3"`
- `remappings.txt` — `rain-math-binary-0.1.1/=dependencies/rain-math-binary-0.1.1/`
  -> `rain-math-binary-0.1.3/=dependencies/rain-math-binary-0.1.3/`
- `soldeer.lock` — regenerated `version` / `url` / `checksum` / `integrity` for
  the `rain-math-binary` entry

There is no `foundry.lock` in this repo; `soldeer.lock` is the only lockfile
Soldeer maintains here.

## Known blocker — this is expected to be red

**Read this before reviewing CI.** The red is upstream, not a defect in this
diff, and it cannot be fixed on this branch.

`rain-math-binary` is listed in `foundry.toml` **solely** to satisfy hardcoded,
version-suffixed transitive imports inside two other Soldeer deps. This repo
sets `[soldeer] recursive_deps = false`, so its flat dep set has to satisfy its
dependencies' own pinned import paths.

Seven files across two published deps pin the `0.1.1` path literal:

```
dependencies/rain-interpreter-interface-0.1.0/src/lib/parse/LibParseMeta.sol
dependencies/rain-interpreter-interface-0.1.0/src/lib/codegen/LibGenParseMeta.sol
dependencies/rain-interpreter-interface-0.1.0/test/src/lib/codegen/LibGenParseMeta.findExpander.t.sol
dependencies/rain-interpreter-interface-0.1.0/test/src/lib/codegen/LibGenParseMetaSlow.sol
dependencies/rainlang-0.1.5/src/lib/op/bitwise/LibOpBitwiseCountOnes.sol
dependencies/rainlang-0.1.5/test/src/lib/op/bitwise/LibOpBitwiseCountOnes.t.sol
dependencies/rainlang-0.1.5/test/src/lib/parse/LibParseState.constantValueBloom.t.sol
```

each containing literally:

```solidity
import {LibCtPop} from "rain-math-binary-0.1.1/src/lib/LibCtPop.sol";
```

Once `remappings.txt` no longer carries a `rain-math-binary-0.1.1/` prefix,
those imports have nothing to resolve against. They are on this repo's compile
path — `script/Build.sol:11` and
`test/concrete/parser/RaindexV6SubParser.pointers.t.sol:17` both
`import {LibGenParseMeta} from "rain-interpreter-interface-0.1.0/src/lib/codegen/LibGenParseMeta.sol"`,
and the subparser reaches `LibParseMeta`. Those files live in gitignored
`dependencies/`, are immutable published Soldeer artifacts, and cannot be
edited from this repo.

Bumping those two deps does **not** unblock it. I pulled the latest published
zips straight from the Soldeer registry and grepped them:

| dependency | version checked | pins |
| --- | --- | --- |
| `rain-interpreter-interface` | 0.1.2 (latest published) | `rain-math-binary-0.1.1` |
| `rainlang` | 0.1.8 (latest published) | `rain-math-binary-0.1.1` |

So there is **no combination of currently published versions** that lets this
repo move off `rain-math-binary-0.1.1`.

The path forward is upstream, in order: `rain-interpreter-interface`
republishes against `rain-math-binary 0.1.3`, then `rainlang` republishes on
top of that. Follow-up commits on this branch will bump those two here once
those versions exist. This PR is the last link in that chain and cannot go
green before them. It is opened to put the blocker on the record with a
reproducible diff.

The alternative — a mismatched alias remapping
(`rain-math-binary-0.1.1/=dependencies/rain-math-binary-0.1.3/`) in
`foundry.toml`'s `remappings` array — defeats the point of the version-suffixed
scheme and was explicitly ruled out for this change.

## QA

- Discriminating tests: n/a - no test was added or changed. The diff is three
  non-source files and adds no Solidity, so there is no new behaviour for a
  test to discriminate. The existing suite is the check that matters, and it
  cannot currently run on this branch (see Known blocker) - so no test can be
  demonstrated failing on base and passing on head.
- Mutations applied: n/a - nothing mutable was written. The upstream
  `0.1.1 -> 0.1.3` delta is NatSpec/comment-only in `src/lib/LibCtPop.sol`,
  which yields no executable line to mutate, and this repo contributes zero new
  statements. Mutating raindex source unrelated to this bump would be probing
  something this PR does not change.
- Oracle: the published Soldeer registry, read independently of this checkout.
  `https://api.soldeer.xyz/api/v1/revision?project_name=rain-math-binary`
  returns 0.1.3 as the newest revision at
  `.../rain-math-binary/0_1_3_19-07-2026_19:05:36_rain.math.zip`, and Soldeer
  recomputed `checksum` / `integrity` itself on download rather than my
  transcribing them. The blocker was established the same way - the published
  `rainlang@0.1.8` and `rain-interpreter-interface@0.1.2` zips were downloaded
  from the registry and grepped directly, not inferred from this working tree.
- Category check: the ask is (A) `foundry.toml` -> 0.1.3, (B) `soldeer update`
  to regenerate the lockfile, (C) `remappings.txt` moved keeping the version
  suffix, (D) zero `rain-math-binary-0.1.1` left in tracked files, (E) build and
  tests green, (F) PR opened. Covered A, B, C, D, F. **E is not covered and
  cannot be** - see the Known blocker; it needs an upstream republish, not a
  change in this repo.

**Verification is delegated to CI, and CI is expected to be RED on this PR**
for the upstream reason set out above — not because of a defect in this diff.
Do not treat that red as something to fix on this branch. Per instruction I did
not run `forge build`, `forge test`, the static/lint gate, the artifact
regeneration or `.gas-snapshot` on this branch. Nothing below is a green claim.

What I actually ran, and its actual output:

- `nix develop .#sol-shell -c forge soldeer install` on unmodified `main` —
  exit 0, `Done installing!`, working tree clean afterwards. Baseline
  dependency install only.
- `nix develop .#sol-shell -c forge soldeer update` after editing
  `foundry.toml` — exit 0, `Updated lockfile` / `Updated remappings` /
  `Done updating!`. Note `soldeer update` **appends** the new remapping without
  pruning the superseded one, and leaves the old `dependencies/` directory in
  place. I deleted `dependencies/rain-math-binary-0.1.1/`, re-ran
  `forge soldeer install` (exit 0), and removed the stale
  `rain-math-binary-0.1.1/` line from `remappings.txt` by hand, so exactly one
  `rain-math-binary` remapping remains.
- Sweep: `git grep -n "rain-math-binary-0.1.1"` over tracked files returns
  **zero hits** (exit 1). A working-tree `grep -rn` excluding `.git`, `out`,
  `cache`, `node_modules` and `target` also returns zero hits outside
  `dependencies/`. The seven remaining occurrences are the published artifacts
  named in the blocker section, which are regenerated by Soldeer and not
  editable here.
- `git status --short` before committing showed exactly `foundry.toml`,
  `remappings.txt`, `soldeer.lock`. The devshell-generated
  `.pre-commit-config.yaml` is `.gitignore`d in this repo and is not in the
  commit.

**No baseline test result exists for `main`, and this PR claims none.** A
baseline `forge build && forge test` on unmodified `main` was started and then
killed before it finished — it reached `Compiling 581 files with Solc 0.8.25`
and produced no pass/fail summary. I therefore have no local evidence about
which failures, if any, are pre-existing on `main`.

Likewise, `forge build` was **not** run on this branch. The blocker above is
established by reading the import statements in the published dependency
sources and this repo's remappings, not by observing a compiler error.

Not run locally, by instruction — flagged as known risk for CI to adjudicate:

- `forge build`, `forge test` (`rainix-sol / test`)
- `slither .`, `forge fmt --check`, `rainix-sol-single-contract`
  (`rainix-sol / static`)
- `script/build-meta.sh`, `forge script script/Build.sol`, `script/build.sh` and
  the `git diff --exit-code` artifact-currency gate (`copy-artifacts`).
  **No generated artifact was regenerated or staged in this PR.** Since the
  library delta is comment-only, no pointer/ABI/meta drift is expected — that
  is a prediction, not a measurement, and `copy-artifacts` is the check that
  settles it.
- `.gas-snapshot` was **not** regenerated. It is checked in, but no rainix
  workflow verifies it (`rainix-sol-static`, `rainix-sol-test` and
  `rainix-copy-artifacts` contain no snapshot check), and a comment-only
  library delta cannot move gas.

## Summary by CodeRabbit

* **Chores**
  * Updated the underlying math component to version 0.1.3.
  * Includes the latest dependency path configuration for improved compatibility and maintenance.


thedavidmeister:
🤖 ai:producer
Design-question: The red is real but the body rules out the one in-repo fix with no human ruling on record.

LOG: both failing checks die on Error (6275): Source "rain-math-binary-0.1.1/src/lib/LibCtPop.sol" not found, imported by dependencies/rain-interpreter-interface-0.1.0/src/lib/parse/LibParseMeta.sol.

Chronology confirms no published upstream helps: rain-math-binary 0.1.3 shipped 2026-07-19, newest rain-interpreter-interface 0.1.2 is 2026-06-15, newest rainlang 0.1.8 is 2026-07-03.

QUESTION: foundry.toml remappings already carries an alias remap for @openzeppelin, so (a) adding "rain-math-binary-0.1.1/=dependencies/rain-math-binary-0.1.3/" is live and is the only in-repo path to green, at the cost of a version-lying literal (the NatSpec-only delta claim is UNVERIFIED); versus (b) hold for an upstream republish nobody has started; versus (c) close as premature. Not blocked-on: no upstream ref exists to type.

thedavidmeister:
🤖 ai:producer
Blocked-on: All three failing checks (rainix-sol/static, rainix-sol/test, copy-artifacts) have ONE cause, read from their logs: Error (6275): Source "rain-math-binary-0.1.1/src/lib/LibCtPop.sol" not found. It is raised only from inside installed dependencies, never from raindex source: rain-interpreter-interface-0.1.0 (LibParseMeta.sol, LibGenParseMeta.sol) and rainlang-0.1.5 (LibOpBitwiseCountOnes.sol). No raindex-specific failure sits on top; compilation never completes, so no test runs.

No published upstream can help: rain-math-binary 0.1.3 shipped 2026-07-19, while newest rain-interpreter-interface 0.1.2 (2026-06-15) and newest rainlang 0.1.8 (2026-07-03) both predate it and still hardcode the 0.1.1 prefix.

Supersedes my earlier ai:design flag on this PR: same upstream cause as the sibling bumps, and the fleet-wide design question is already asked once on rain.merkle#16, so it routes here instead of being repeated.
blocked-by rainlanguage/rainlang#551

PR #2838, opened 2026-08-22
Migrate release machinery to rain-deploy 0.1.7 tag-driven releases

thedavidmeister:
Replaces merge-driven Soldeer publishing with the org's tag-release machinery: releases now cut from `sol-v*` tags via `rainix-tag-release.yaml@main`, on rain-deploy 0.1.7's `BuildScript`/candidate-snapshot flow. `RaindexDeploySuites` declares the six release candidates (keys match the `manual-sol-artifacts.yaml` choice list); `LibRaindexDeploy` stays hand-written, repointed at `src/generated/candidate/`; `run()` remains a read-only determinism check and only `cutRelease()` freezes. Deps: forge-std 1.16.2, rain-sol-codegen 0.1.36, rain-deploy 0.1.7, with three alias remappings in `foundry.toml` for transitive pins on `forge-std-1.16.1/`, `rain-sol-codegen-0.1.0/`, `rain-deploy-0.1.2/` (Soldeer allows one version per package name; the codegen interface delta is `pure`→`view` only, and the rain-deploy alias is test-only).

## No pin moved (verified)

The regenerated `src/generated/candidate/*` constants are byte-identical to the previous `src/generated/*.pointers.sol` for all six contracts: `BYTECODE_HASH`, `DEPLOYED_ADDRESS`, `RUNTIME_CODE` for all six, plus `CREATION_CODE` for the two that had it (RaindexV6, RaindexV6SubParser). Verified by extracting every one of those constant declarations from `git show <parent>:src/generated/<Name>.pointers.sol` and from `src/generated/candidate/<Name>.sol` and comparing per contract (whitespace-normalized sha256 over the concatenated declarations) — six of six identical. The first Build run after the migration transiently showed the three arb contracts differing: a bootstrap artifact — the stub candidate files held `DEPLOYED_ADDRESS = address(0)` and the arb contracts embed `LibRaindexDeploy.RAINDEX_DEPLOYED_ADDRESS`/`ROUTE_PROCESSOR_DEPLOYED_ADDRESS` in their bytecode. With real constants in place the next run converged, and a further run plus `forge fmt` is a byte-for-byte fixpoint.

## Expected red: `frozen-snapshots-append-only`

Deleting the ten `src/generated/<tag>/` dirs (`0_1_4`…`0_1_13`) will fail rainix-sol-static's `frozen-snapshots-append-only` gate, because `src/generated/` still exists in HEAD (`candidate/`). The deletion is deliberate and ruled — migrations carry the current shape forward only, no historicals — and the historicals are not being kept just to keep that gate green. Merging over the red (or an upstream rainix exemption) is the human's call.

## First tag will fail the 7-network verify

rain-deploy 0.1.7's `RainDeployVerifyChain` reads `LibRainDeploy.supportedNetworks()` — a fixed set of 7 networks including `ethereum` and `hyperevm` — with no override point. Raindex is live on arbitrum/base/flare/polygon (+ base_sepolia) only. Harmless in this PR (the released record is empty, and the check early-returns on length 0), but the first `sol-v0.1.14` tag will fail its verify step unless raindex is deployed to ethereum + hyperevm first, or rain.deploy grows a per-repo network set. `Deploy.sol` meanwhile pins `deployNetworks()` to the current five, exactly rain-deploy 0.1.2's `supportedNetworks()`, so deploy behaviour is unchanged.

## End of merge-driven publishing

Moving off `rainix-autopublish` means merges to main no longer publish `raindex` to Soldeer — only `sol-v*` tags do. Anything consuming `raindex` from the registry expecting per-merge patch bumps loses that.

## DescribeSubParser: one explicit network

`RainDeployBroadcast.run()` is external, non-virtual, and has no post-deploy hook, so the subparser's `LibDescribedByMeta.emitForDescribedAddress` metaboard emission moved out of `Deploy.sol` into `script/DescribeSubParser.sol`, dispatched through `manual-sol-artifacts.yaml`'s new `script`/`network` inputs. It emits on exactly one network per dispatch, named explicitly by `DEPLOYMENT_NETWORK`. The old in-line version also hit only one network — whichever fork `deployAndBroadcast` happened to leave selected last — so this makes the one-network behaviour deliberate instead of accidental; it does not loop all networks.

## Registry-sweep property

With the historicals gone, published Soldeer versions 0.1.4–0.1.13 no longer have their frozen `src/generated/<tag>/` records in-repo. The "every published version keeps its frozen record" property is tracked upstream as rainlanguage/rainix#355 — referenced, not reimplemented here.

## QA

- Discriminating tests: `test/abstract/RaindexDeployChain.t.sol` + `test/abstract/RaindexDeploySnapshot.t.sol` (bindings of rain-deploy 0.1.7's `RainDeployVerifyChain`/`RainDeployVerifySnapshot` suites) — each fails on base by non-compilation (verified by inspection: base's `soldeer.lock` pins rain-deploy 0.1.2 and has no `dependencies/rain-deploy-0.1.7/` or `src/generated/candidate/` for their imports to resolve). No behavioural test discriminates beyond that: this is release-machinery migration and deliberately moves no contract bytecode.
- Mutations applied: n/a — no mutation campaign was run for this change; the change's single invariant (no pin moved) is checked by direct byte-comparison of constants against the parent commit, not by a test, and this PR intentionally commits no mutants config.
- Oracle: the frozen pointer constants in the parent commit (`git show 64dddfd:src/generated/<Name>.pointers.sol`) are the expected values for every regenerated `candidate/` constant, independent of the regeneration that produced them; the machinery shape follows rain.factory.deploy / rain.math.float.deploy as org precedent.
- Category check: no linked issue; the work order is the tag-release migration handoff — covered: package/deps (`foundry.toml`, `soldeer.lock`), suites declaration (`RaindexDeploySuites`), scripts (`Build`, `Deploy`, `DescribeSubParser`), generated-tree swap (`candidate/` in, historicals out), workflows (`package-release`, `manual-sol-artifacts`), docs (`CLAUDE.md`).


thedavidmeister:
## Current CI state

`rainix-sol / static` and `rs-static` are red — but NOT yet for the append-only reason given in the PR body. Both fail at rainix's `agent-context-cap` gate (CLAUDE.md + its `@AGENTS.md` import = 12547 bytes vs the 4096 cap), which is red on every recent main commit — pre-existing, not introduced here. In `rainix-sol / static` that step fails before `frozen-snapshots-append-only` runs, so the ruled append-only red is currently masked (the step shows as skipped). Once the context cap is cut, the static job will fail at `frozen-snapshots-append-only` instead — the deliberate red described in the body.

A cap cut is prepared in the working clone but not committed (CLAUDE.md + AGENTS.md trimmed to 4041 charged bytes; the path-scoped rulings — CI workflow conventions, Rust functional style, committed derived artifacts — moved to `.claude/rules/*.md` with `paths:` frontmatter, which the gate does not charge; discoverable content cut per the gate's own criteria). The operator's permission system refused an agent commit touching agent-config files, so landing that cut is the human's move.

Everything else is green, including `rainix-sol / test` (fork suite) and `copy-artifacts` (Build determinism / clean-tree check).

thedavidmeister:
Correction to the note above, one detail that matters for landing the cap cut: raindex's `.gitignore` line 22 ignores `.claude/` wholesale, so the relocated `paths:`-scoped rules files under `.claude/rules/` would NOT be picked up by a plain `git add -A` — a naive commit of the prepared cut would ship only the trimmed CLAUDE.md/AGENTS.md and silently drop the relocated rulings (CI workflow conventions, Rust functional style, committed derived artifacts). Landing the cut therefore also needs the `.gitignore` entry narrowed (e.g. keep ignoring `.claude/settings.local.json` but track `.claude/rules/`), or the rulings kept somewhere tracked. All three prepared files sit in the working clone for whoever lands it.

PR #2841, opened 2026-08-23
Cut release 0.1.14: freeze candidate pins, target the standard seven networks

thedavidmeister:
The repo's first tag-driven release under the #2838 machinery, plus the standards conformance that release surfaced.

- `foundry.toml` `[external.package].version` 0.1.13 → 0.1.14, in lockstep with the frozen record it names.
- `forge script ./script/Build.sol --sig 'cutRelease()'` output committed verbatim: `src/generated/0_1_14/` (byte-identical copy of `candidate/`, six of six verified) and the released-suites libs regenerated from the record that now holds 0.1.14.
- `Deploy.sol` drops its five-network `deployNetworks()` override: raindex targets the full `LibRainDeploy.supportedNetworks()` seven — arbitrum, base, base_sepolia, ethereum, flare, hyperevm, polygon — via the inherited default. Script-only; no contract bytecode moves.
- Residue: `build-start-blocks.sh` read the deleted `src/generated/RaindexV6.pointers.sol` (broken since #2838) and now reads `src/generated/candidate/RaindexV6.sol`; three test-util comments and the CLAUDE.md chain count updated to match.

**Deploy state, and why it gates the tag.** Freezing 0.1.14 arms `RainDeployVerifyChain` over the released record, so the tag's fork suite asserts the frozen suite is live on every network in the set. Verified by `eth_getCode` against the candidate `RaindexV6` (`0x37FC0EFec37D19f8A221aa4F8F7600C9ba2AcD20`): live on arbitrum, base, base_sepolia, flare, polygon; **absent on ethereum and hyperevm**. Those two need broadcasting before `sol-v0.1.14` is pushed, or the tag run goes red at the chain suite. HyperEVM needs the big-blocks path (the deployer already has big blocks on).

Merging this PR publishes nothing — publication happens only when the `sol-v0.1.14` tag is pushed, which re-runs the fork suite and release-guard first.

## QA

- Discriminating tests: `test/abstract/RaindexDeployChain.t.sol` and `test/abstract/RaindexDeploySnapshot.t.sol` become discriminating *because of* this commit — before it, `releasedSuites()` was empty and both took their `length == 0` early return, asserting nothing; after it they check the 0.1.14 record's pins for internal consistency and on-chain presence. The chain half is expected red for ethereum/hyperevm until those deploys run (see above), which is the check doing its job rather than a defect in the diff.
- Mutations applied: n/a — no logic in the diff. The freeze is generated output committed verbatim plus a script-only network-set change; the one invariant that matters (no pinned value moved) is verified by byte-comparison of `0_1_14/` against `candidate/`, not by a mutant.
- Oracle: the `candidate/` constants at the parent commit are the expected values for every frozen constant (six of six byte-identical); the seven-network set comes from `LibRainDeploy.supportedNetworks()` in rain-deploy 0.1.7, not from this repo's prior override; on-chain presence comes from `eth_getCode` against live RPCs, independent of anything in the tree.
- Category check: work order was "cut the release" plus "get raindex up to latest deploy standards" — covered: version bump, frozen record, released-suites regeneration, network set corrected to the standard seven, post-#2838 residue repaired. Deferred with reason: the library/deploy repo split (#552-shaped), which the user ruled out of scope for raindex.

## Summary by CodeRabbit

* **New Features**
  * Added release 0.1.14 deployment artifacts for Raindex V6 and related trading components.
  * Deployments now broadcast across all supported chains, rather than a limited subset.
  * Added support for the latest route processing, order-taking, flash-borrowing, and subparser components.

* **Documentation**
  * Updated deployment and testing guidance to reflect the current generated release artifacts and seven-chain deployment coverage.


thedavidmeister:
## Deploy-repo standard audit

Audited this branch against the org's current deploy-repo standard, derived from
`rain.deploy` at `sol-v0.1.7`/`main` (`BuildScript`, `RainDeploySuitesBase`,
`RainDeployVerifySnapshot`, `RainDeployVerifyChain`, `RainDeployBroadcast`,
`LibRainDeploy.supportedNetworks`) and the four conforming exemplars:
`rain.math.float.deploy`, `rain.extrospection.deploy`, `rain.factory.deploy`,
`rain.metadata.deploy`.

### Already conforming on this branch

Layout (`src/abstract/RainDeploySuitesBase.sol` re-export shim byte-identical to
all four exemplars; `src/generated/candidate/` + append-only `src/generated/0_1_14/`,
six each; six per-contract `Lib*Released.sol` plus the generated
`LibReleasedSuites.sol` aggregate) · `script/Build.sol` shape · `script/Deploy.sol`
as the one-line `contract Deploy is RaindexDeploySuites, RainDeployBroadcast {}`
with no `deployNetworks()` override · `foundry.toml` (`[external.package]`
name/version in lockstep with the frozen dir, `solc`/`optimizer`/`evm_version`/
`bytecode_hash`/`cbor_metadata` pins, `fs_permissions` covering `./src` and
`foundry.toml`, `recursive_deps = false`, `forge-std 1.16.2` + `rain-deploy 0.1.7`
+ `rain-sol-codegen 0.1.36`, all **seven** `[rpc_endpoints]` and `[etherscan]`
entries with `chain = 1` on ethereum and `chain = 999` on hyperevm) ·
`package-release.yaml` with **no** `snapshot-generate-cmd` · `rainix-sol.yaml` ·
`copy-artifacts.yaml` on `rainix-copy-artifacts@main` · `manual-sol-artifacts.yaml`
`workflow_dispatch`-only with a `choice` list matching all six declared candidate
keys · both verify abstracts bound in separate contracts · `.soldeerignore` ·
`CLAUDE.md` + `@AGENTS.md` at 4042 of the 4096-byte cap.

`optimizer_runs = 800` is a documented load-bearing exception (EIP-170), the same
class as `rain.math.float.deploy`'s `1000000`. Not touched — it moves every address.

### Fixed here

- **`slither.config.json`** now filters
  `src/abstract/(RaindexDeploySuites|RainDeploySuitesBase)\.sol`, which all four
  exemplars carry and this repo did not. This is what `rainix-sol / static` was
  red on: all 20 findings were `too-many-digits`/`dead-code` on the suite
  declaration and `uninitialized-local` inside the packaged base. Appended, not
  replaced, so nothing previously reported stops being reported. **Static is now
  green.**
- **`test/script/Build.t.sol` + `test/util/concrete/BuildHarness.sol`.**
  `Build.generatedContracts()` and `RaindexDeploySuites.candidateSuites()` are two
  hard-coded arrays in two files with nothing tying them together; a candidate
  dropped from the generator still compiles, and `cutRelease()` then omits it from
  the release permanently, invisibly to both the frozen-record walk and the chain
  check. Verified as a real gate: dropping `arb-generic-pool-flash-borrower` from
  the generator fails the set-equality test with `5 != 6`; unmutated it is 4/4.
- **`Manual sol artifacts` exposes `legacy`** (default false). Dropping the
  five-network override put HyperEVM in the set, and HyperEVM's RPC rejects the
  fee-history ranges EIP-1559 estimation asks for.
- **`RaindexDeploySuites.sol` pragma floated** to `^0.8.25`. It was the only
  abstract in this repo pinned with `=`, and `src/` is published.

### Blocking, and NOT worked around

`rainix-sol / test` is red on one test and it is a true fact about the world:

```
[FAIL: NotDeployedOnNetwork("ethereum", "raindex@0_1_14", 0x37FC0EFec37D19f8A221aa4F8F7600C9ba2AcD20)]
```

694 of 695 pass. Confirmed on chain: the address is live on arbitrum, base and
flare, and holds **no code on ethereum or hyperevm** — raindex has never deployed
there, which is exactly what dropping the five-network override now surfaces.

This is step 1 of the standard's three-step order (deploy → verify → tag) and it
is key custody and real money, so it needs a human dispatch, not a code change.
To clear it: dispatch `Manual sol artifacts` once per suite, **serialized** (one
deployer EOA, they race the nonce), for all six suites. The broadcast is
idempotent, so the five chains that already have the code are skipped.

Two things to know before dispatching HyperEVM:

- Set `legacy: true` on the dispatch (that input is what this PR adds).
- `RaindexV6`'s creation code is **24,044 bytes** and `RouteProcessor4`'s is
  16,301, so the code-deposit gas is far over HyperEVM's small-block cap. The
  deployer must be a HyperCore user and toggled into big blocks first, via
  `rain.deploy`'s `Manual credit hypercore` and `Manual big blocks` workflows
  (both admin-gated). The big-blocks flag is persistent per address, so if this
  deployer was already toggled it still is.

### Deviations left alone, with reason

- **`copy-artifacts.yaml` / `name: copy-artifacts`** — the newest exemplar
  (`rain.metadata.deploy`) calls this `git-clean.yaml` / `Git is clean`. Same
  reusable, same behaviour; renaming a workflow renames a status check, which is
  branch-protection surface and does not belong in a release cut.
- **`remappings.txt` committed** — the four exemplars gitignore it. It is
  currently consistent (`copy-artifacts` is green), and removing it risks a job
  that never runs `forge soldeer install`.
- **`test/abstract/…` rather than `test/src/abstract/…`** — repo-wide convention
  here, not enforced by `rainix-static`. Conforming means renaming the whole test
  tree.
- **36 of this repo's own `.sol` files import `rain-deploy-0.1.2/…`**, resolved
  by the `foundry.toml` alias onto the installed 0.1.7. The alias has to stay for
  `rainlang` 0.1.5, but this repo spelling a version it does not pin defeats the
  point of version-qualified imports. A mechanical 36-file rename; worth its own
  PR, not this one.
- No `<Thing>DeploySuitesTest` (key/artifactPath/dependency pins) — recommended
  but only one of four exemplars has it.
- `AGENTS.md` points at `.claude/rules/`, which does not exist here. Stale, and it
  is charged against the 4096-byte cap, which has 54 bytes of headroom.

PR #2843, opened 2026-08-23
Take the MetaBoard dep from rain-metadata-deploy, not the split-away pin

thedavidmeister:
## What blocked the deploy

`RaindexDeploySuites.raindexDependencies()` declared

```solidity
deps[2] = LibMetaBoardDeploy.METABOARD_DEPLOYED_ADDRESS;
```

through `rain-metadata` **0.1.0**, whose `src/lib/deploy/LibMetaBoardDeploy.sol`
names the pre-split v1 MetaBoard **`0xfb8437Ae…6928`**. That board has code on
arbitrum, base, base_sepolia, flare and polygon — but **not** on ethereum or
hyperevm. `RainDeployBroadcast` checks declared dependencies before it
broadcasts, so every dispatch died on:

```
MissingDependency("ethereum", 0xfb8437Ae…6928)
```

## What this does

The MetaBoard has split out of `rain.metadata` into `rain.metadata.deploy`.
Its published **`rain-metadata-deploy` 0.1.0** (soldeer, 2026-08-21, tag
`sol-v0.1.0`) pins **`0x8fD50fF9Db9835ba1B61394752A26F53D721D2a1`**, which is
live on all seven supported networks. The declared dependency moves there.

| pin | before | after | why |
|---|---|---|---|
| `rain-metadata` | 0.1.0 | **0.1.7** | still needed for `LibMeta`, `LibDescribedByMeta`, `IMetaV1_2`, `IMetaBoardV1_2`, `IDescribedByMetaV1` |
| `rain-metadata-deploy` | — | **0.1.0** | new; the only source of `LibMetaBoardDeploy` |

Both are the newest published revisions on the soldeer registry at the time of
writing; `rain.metadata.deploy` `sol-v0.1.0…main` touches no `src/*.sol`, so a
follow-up 0.1.1 there would not move this address.

### The `rain-metadata` bump is API-safe

`diff -rq` of the two published packages' `src/` trees: every file this repo
imports is **byte-identical** between 0.1.0 and 0.1.7. What 0.1.7 removes is
exactly `src/concrete/MetaBoard.sol` and `src/lib/deploy/` (the split), and what
it adds is `src/lib/LibIMetaBoardV1_2.sol`. No behaviour this repo reaches
changed.

### One alias remapping

`rainlang` 0.1.5's `BaseRainlangSubParser` imports `IDescribedByMetaV1` under
the literal `rain-metadata-0.1.0/` prefix, and soldeer installs one version per
package name, so `foundry.toml` aliases the old prefix onto the 0.1.7 install:

```toml
"rain-metadata-0.1.0/=dependencies/rain-metadata-0.1.7/",
```

Same pattern the repo already uses for `forge-std-1.16.1/`,
`rain-sol-codegen-0.1.0/` and `rain-deploy-0.1.2/`. `IDescribedByMetaV1.sol` is
byte-identical across the two versions, and the alias is what keeps it ONE type
— two installed copies would be two distinct `IDescribedByMetaV1`s and the
sub-parser would stop satisfying this repo's own ERC-165 / interface checks.

## Does anything move? No — only `DEPENDENCIES`

The declared dependency list is recorded in each snapshot's `DEPENDENCIES`
constant, so the rolling candidate had to be regenerated. Only `RaindexV6` and
`RaindexV6SubParser` carry `raindexDependencies()`; the other four declare
`new address[](0)` and did not change at all.

**Verified, per contract, before vs after regeneration:**

| contract | `DEPLOYED_ADDRESS` | `BYTECODE_HASH` | `CREATION_CODE` | `RUNTIME_CODE` |
|---|---|---|---|---|
| RaindexV6 | `0x37FC0EFec37D19f8A221aa4F8F7600C9ba2AcD20` unchanged | unchanged | unchanged | unchanged |
| RaindexV6SubParser | `0x09Bc7AF266012F44fb41D8Bd682da931666605e1` unchanged | unchanged | unchanged | unchanged |
| RouteProcessor4 | `0x6E2d0e71d900474b262E545Bc4C98b71ab368d21` unchanged | unchanged | unchanged | unchanged |
| GenericPoolRaindexV6ArbOrderTaker | `0xE84c106B0A89A164d2D65205B9EBAE37c15Fd84a` unchanged | unchanged | unchanged | unchanged |
| RouteProcessorRaindexV6ArbOrderTaker | `0x1350420cbf3E9eb8F1734bbe466e0F303579eE24` unchanged | unchanged | unchanged | unchanged |
| GenericPoolRaindexV6FlashBorrower | `0x032d9D94A79909F3b337ECFE6f73f4e86bA79c7E` unchanged | unchanged | unchanged | unchanged |

The whole diff under `src/generated/candidate/` is two lines — the third address
inside `DEPENDENCIES`:

```
-…000000000000000000000000fb8437aefbb8031064e274527c5fc08e30ac6928
+…0000000000000000000000008fd50ff9db9835ba1b61394752a26f53d721d2a1
```

That is expected: the metaboard address is a *precondition* the broadcast gate
checks, never a constructor argument or a linked constant, so it was never in
the bytecode. **Nothing needs redeploying that did not already.** The addresses
`src/generated/0_1_14/` names are the addresses this PR still names.

## Why the freeze is re-cut as 0.1.15, not edited in place

Regenerating the candidate puts `src/generated/0_1_14/` out of step with a fresh
regen, which `release-guard` check (4b) refuses at tag time. 0.1.14 is merged to
main but **never tagged or published** — soldeer's newest `raindex` is 0.1.13 —
so the freeze does need re-cutting.

It could not be re-cut *in place*. Two independent guards forbid it:

1. `rain-deploy`'s own `freeze()` refuses. Run against this tree it gives, and I
   ran it:
   ```
   Error: script failed: SnapshotAlreadyFrozen("0_1_14", "src/generated/0_1_14")
   ```
   Deleting the dir first is not an escape either: `LibRaindexV6Released.sol`
   imports `../generated/0_1_14/RaindexV6.sol`, so removing it makes the tree
   uncompilable and the generator that would recreate it cannot run — the exact
   case `release-guard`'s docs call out for not re-freezing.
2. `frozen-snapshots-append-only` (rainix-sol / static) reds any branch that
   modifies or deletes a file under an existing `<tag>/`.

So `[external.package].version` goes 0.1.14 → **0.1.15** and `cutRelease()`
appends `src/generated/0_1_15/`, byte-identical to the freshly regenerated
`candidate/`. `src/generated/0_1_14/` is untouched and keeps its old dependency
list — which is what an append-only record is for: a release records what it
required, and `LibRaindexV6Released` reads each entry's dependencies from its
own frozen file. Version 0.1.14 is simply skipped on soldeer; the next tag is
`sol-v0.1.15`.

If a reviewer would rather burn no version number, the alternative is to hand-
overwrite `src/generated/0_1_14/` and accept a red `rainix-sol / static` — I did
not take that, and I have not touched the gate.

## Verification run locally

- `forge soldeer install` / `update` — lockfile and `remappings.txt` regenerated;
  no other dependency version moved.
- `forge script ./script/Build.sol` — clean, and **idempotent**: a second run
  produces no further diff.
- `./script/build-meta.sh`, `forge build`, `forge fmt` — no drift, so
  `copy-artifacts` has nothing to report.
- `forge test --match-contract 'RaindexDeploySnapshotTest|BuildTest'` — 7/7 pass,
  including `testSnapshotMatchesSource`, `testSnapshotInternallyConsistent` and
  `testEveryFrozenSnapshotIsReleased` (which now matches both `0_1_14` and
  `0_1_15` by derived address).

## Expected red

`RaindexDeployChainTest` / `testProdDeploy*` stay red until the deploys actually
run — the released suites are not on chain yet. That is the blocker this PR
exists to clear, not a regression, and it is deliberately not weakened here.
## The dependency gate, verified against live chains

`eth_getCode` for all three entries of `raindexDependencies()`, on all seven
supported networks, run against public RPCs while writing this:

| network | log tables `0xc51a1425…f5f5` | TOFU decimals `0x200e12d1…b9389` | MetaBoard `0x8fD50fF9…D2a1` | old MetaBoard `0xfb8437Ae…6928` |
|---|---|---|---|---|
| arbitrum | code | code | code | code |
| base | code | code | code | code |
| base_sepolia | code | code | code | code |
| ethereum | code | code | **code** | **`0x`** |
| flare | code | code | code | code |
| hyperevm | code | code | **code** | **`0x`** |
| polygon | code | code | code | code |

The two `0x` cells are exactly the `MissingDependency` this PR removes. On
ethereum and hyperevm the runtime returned at `0x8fD50fF9…D2a1` hashes to
`0xb7b74d879e97f272145d32ce18bd060f1e9f387696911376191ca10ecf24bed6`, which is
the `BYTECODE_HASH` in `rain-metadata-deploy` 0.1.0's own snapshot — so it is
the right board, not just some contract at the right address.

## Still blocking the deploy

Nothing in this repo, and nothing on chain. Once this merges, dispatching
`manual-sol-artifacts.yaml` per suite (serialized — one deployer EOA) should
clear the dependency gate on all seven networks. After the deploys land,
`RaindexDeployChainTest` goes green and `sol-v0.1.15` can be tagged.


thedavidmeister:
@coderabbitai — on the one flagged risk ("the release version should be confirmed against the registry so the generated snapshot and release tags do not conflict with an existing published version"): checked, and 0.1.15 is free.

Soldeer's newest published `raindex` revision is **0.1.13** (`GET https://api.soldeer.xyz/api/v1/revision?project_name=raindex` returns 0.1.13, 0.1.12, 0.1.11, 0.1.10, 0.1.9 …). **0.1.14 was never published** — it was frozen as `src/generated/0_1_14/` and merged to main, but the `sol-v0.1.14` tag was never pushed, because the deploy it was cut for is the one this PR unblocks. So no published revision, and no existing `sol-v*` tag, collides with 0.1.15.

That is also why the freeze is appended as 0_1_15 rather than 0_1_14 being edited: `cutRelease()` refuses to overwrite an existing tag dir (`SnapshotAlreadyFrozen("0_1_14", "src/generated/0_1_14")` — I ran it), and `frozen-snapshots-append-only` reds a modified snapshot. Both guards stayed on; `rainix-sol / static` is green.

For the record, the new MetaBoard pin is verified live rather than assumed. `eth_getCode` at `0x8fD50fF9Db9835ba1B61394752A26F53D721D2a1` returns code on all seven supported networks, and on ethereum and hyperevm — the two that were failing — the returned runtime hashes to `0xb7b74d879e97f272145d32ce18bd060f1e9f387696911376191ca10ecf24bed6`, which is exactly the `BYTECODE_HASH` in `rain-metadata-deploy` 0.1.0's snapshot. The old `0xfb8437Ae…6928` board returns `0x` on both of those chains, which is the `MissingDependency` this PR fixes.

thedavidmeister:
## CI status: every check green except the one that is supposed to be red

15 of 16 checks pass, including all six required by the `main` ruleset (`test` ×2, `subgraph-test`, `test-js-bindings`, `wasm-artifacts`, `wasm-browser-test`, `wasm-test`), plus `copy-artifacts` and all three `rainix-sol` jobs bar one.

The single failure is `rainix-sol / test / test`, and it is one test:

```
[FAIL: NotDeployedOnNetwork("ethereum", "raindex@0_1_14", 0x37FC0EFec37D19f8A221aa4F8F7600C9ba2AcD20)]
  testSuitesLiveOnEverySupportedNetwork()
  test/abstract/RaindexDeployChain.t.sol:RaindexDeployChainTest
```

**This is pre-existing, not introduced here.** The same job on this PR's base commit (`5984fb5`, run [32624001121](https://github.com/rainlanguage/raindex/actions/runs/32624001121)) fails with the byte-identical assertion — same test, same network, same suite key, same address. RaindexV6 simply is not on ethereum yet, which is the deploy this PR unblocks. It has deliberately not been weakened, skipped or filtered.

Worth noting the failing suite key is `raindex@0_1_14`, not `0_1_15`: the test fails on the first missing derivation and 0_1_14 sorts first. Both releases name the same address, so there is one deployment to do, not two.

Note also that `rainix-sol / static / static` **passes** — that is the job carrying `frozen-snapshots-append-only`, confirming `src/generated/0_1_14/` was left untouched and 0_1_15 was appended rather than the frozen record being rewritten.

Sequence from here: merge → dispatch `manual-sol-artifacts.yaml` once per suite, serialized → `RaindexDeployChainTest` goes green → tag `sol-v0.1.15`. Not merging this myself.

thedavidmeister:
Bumped `rain-metadata-deploy` 0.1.0 → 0.1.1 in 248bbfd — a routine bump to the latest published pin, now that `sol-v0.1.1` is tagged and on the registry (`GET /api/v1/revision?project_name=rain-metadata-deploy` lists 0.1.1, published 2026-08-23T08:23:27Z).

**Nothing generated moved.** 0.1.1 rebuilds the package against `rain-metadata` 0.1.7 — the version this PR already moved raindex to — so the two are no longer built against different `rain-metadata` trees. The MetaBoard pin itself is unchanged: the package's own `0_1_0/`, `0_1_1/` and `candidate/` snapshots of `MetaBoard.sol` are byte-identical (`diff` clean), keeping `DEPLOYED_ADDRESS` at `0x8fD50fF9Db9835ba1B61394752A26F53D721D2a1` and `BYTECODE_HASH` at `0xb7b74d879e97f272145d32ce18bd060f1e9f387696911376191ca10ecf24bed6`.

That matters for the live-chain check in my previous comment: the runtime hash I verified on all seven networks was quoted against 0.1.0's snapshot, and it is the same constant in 0.1.1, so that verification carries over unchanged rather than needing a re-run.

Because `raindexDependencies()[2]` names that address, every snapshot's `DEPENDENCIES` constant is untouched, and so are `BYTECODE_HASH`, `DEPLOYED_ADDRESS`, `CREATION_CODE` and `RUNTIME_CODE` for all six candidates. Confirmed by a from-scratch `forge clean && forge script ./script/Build.sol && forge fmt`: the tree comes back clean (only the five hand-edited pin files differ), and `src/generated/0_1_15/` is byte-identical to the freshly regenerated `src/generated/candidate/`. `src/generated/0_1_14/` is untouched. Nothing needs redeploying that did not already.

The version is spelled in the import prefix, so the bump is the soldeer pin, the remapping and the two `LibMetaBoardDeploy` call sites (`RaindexDeploySuites`, `script/DescribeSubParser.sol`). No alias remapping was needed — unlike `rain-metadata-0.1.0/`, nothing imports `rain-metadata-deploy` under a stale prefix. 0.1.1 also stopped shipping its rust crate (package 24209 bytes vs 56613, solidity only); nothing here consumed it.

`rainix-sol / test` stays red on the same single test as before — `RaindexDeployChain.t.sol:testSuitesLiveOnEverySupportedNetwork`, `NotDeployedOnNetwork("ethereum", "raindex@0_1_14", 0x37FC0EFec37D19f8A221aa4F8F7600C9ba2AcD20)`, 694 passing. Unchanged by this bump and expected until the deploys this PR unblocks are run.

PR #2845, opened 2026-08-25
feat: add Raindex market SDK

findolor:
## Summary

- Add a product-neutral market catalog to the Rust SDK.
- Discover direct quote-token pairs from active Raindex orders through the existing local-database/subgraph query routing.
- Compute native/server-side order books, rolling 24-hour statistics, and recent trades, while using registry token entries only for metadata and variant normalization.

## Key decisions

- Active orders determine which markets exist. The registry does not need to list every indexed token and marketListed is not used.
- marketQuote identifies the quote token for a network; no stock category, symbol, ISIN, USDC address, or application-specific behavior is hardcoded.
- Discovery performs two filtered active-order reads per network: quote token in inputs and quote token in outputs. It does not scan the complete order dataset.
- Indexed name, symbol, and decimals are the fallback for tokens absent from the registry. Registry metadata enriches known tokens and canonicalizes declared unwrapped or legacy variants.
- Route reads through the ready local database when available and through the configured subgraph otherwise. There is no strict or partial multi-subgraph policy.
- Treat Raindex as an orderbook DEX: ticker IDs use contract addresses and executable orders provide depth, with no AMM pool ID or liquidity formula.
- Fetch the 24-hour trade candidate set once per chain and normalize only quote/base executions in memory.
- Perform ERC4626 ratio reads only for tokens that declare an unwrappedAddress or legacyAddress, and cycle configured RPCs with bounded per-attempt deadlines.
- Keep the full market snapshot native/server-side. The web page consumes the cached REST API so external consumers and the website display identical data.

## Validation

- cargo fmt --all -- --check
- git diff --check
- cargo test -p raindex_common --lib: 1,202 passed
- cargo check -p raindex_common --target wasm32-unknown-unknown
- cargo clippy -p raindex_common --lib --all-features -- -D warnings -D clippy::all
- npm run build -w @rainlanguage/raindex
- npm run test -w @rainlanguage/raindex: 124 passed

### DIFF
--- a/crates/common/src/raindex_client/markets/catalog.rs
+++ b/crates/common/src/raindex_client/markets/catalog.rs
@@ -1,0 +1,681 @@
+use super::*;
+use crate::raindex_client::vaults::RaindexVaultToken;
+use std::collections::BTreeSet;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct IndexedToken {
+    address: Address,
+    name: Option<String>,
+    symbol: Option<String>,
+    decimals: u8,
+}
+
+impl From<RaindexVaultToken> for IndexedToken {
+    fn from(token: RaindexVaultToken) -> Self {
+        Self {
+            address: token.raw_address(),
+            name: token.raw_name().map(ToOwned::to_owned),
+            symbol: token.raw_symbol().map(ToOwned::to_owned),
+            decimals: token.raw_decimals(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct IndexedOrder {
+    chain_id: u32,
+    raindex: Address,
+    inputs: Vec<IndexedToken>,
+    outputs: Vec<IndexedToken>,
+}
+
+impl From<RaindexOrder> for IndexedOrder {
+    fn from(order: RaindexOrder) -> Self {
+        Self {
+            chain_id: order.chain_id(),
+            raindex: order.raw_raindex(),
+            inputs: order
+                .input_vaults()
+                .iter()
+                .map(|vault| vault.token().into())
+                .collect(),
+            outputs: order
+                .output_vaults()
+                .iter()
+                .map(|vault| vault.token().into())
+                .collect(),
+        }
+    }
+}
+
+struct MarketAccumulator {
+    base: RaindexMarketToken,
+    quote: RaindexMarketToken,
+    raindex_addresses: BTreeSet<Address>,
+}
+
+pub(super) async fn discover_markets(
+    client: &RaindexClient,
+    options: &MarketListOptions,
+) -> Result<Vec<RaindexMarket>, RaindexError> {
+    let registry_tokens = client.get_all_tokens()?.into_values().collect::<Vec<_>>();
+    let quotes = quote_tokens(&registry_tokens, options)?;
+    if quotes.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    let chain_ids = quotes.keys().copied().collect::<Vec<_>>();
+    let quote_addresses = quotes
+        .values()
+        .map(|quote| quote.address)
+        .collect::<Vec<_>>();
+    let orders = fetch_active_orders(client, chain_ids, quote_addresses).await?;
+    markets_from_indexed_orders(
+        orders.into_iter().map(IndexedOrder::from),
+        &registry_tokens,
+        &quotes,
+        options,
+    )
+}
+
+async fn fetch_active_orders(
+    client: &RaindexClient,
+    chain_ids: Vec<u32>,
+    quote_addresses: Vec<Address>,
+) -> Result<Vec<RaindexOrder>, RaindexError> {
+    let mut orders = Vec::new();
+    for filters in active_order_filters(quote_addresses) {
+        orders.extend(fetch_market_orders_paginated(client, chain_ids.clone(), filters).await?);
+    }
+    Ok(orders)
+}
+
+fn active_order_filters(quote_addresses: Vec<Address>) -> [GetOrdersFilters; 2] {
+    [
+        GetOrdersFilters {
+            active: Some(true),
+            tokens: Some(GetOrdersTokenFilter {
+                inputs: Some(quote_addresses.clone()),
+                outputs: None,
+            }),
+            ..Default::default()
+        },
+        GetOrdersFilters {
+            active: Some(true),
+            tokens: Some(GetOrdersTokenFilter {
+                inputs: None,
+                outputs: Some(quote_addresses),
+            }),
+            ..Default::default()
+        },
+    ]
+}
+
+fn quote_tokens(
+    tokens: &[TokenCfg],
+    options: &MarketListOptions,
+) -> Result<BTreeMap<u32, TokenCfg>, RaindexError> {
+    let selected_chains = options
+        .chain_ids
+        .as_ref()
+        .map(|ids| ids.iter().copied().collect::<HashSet<_>>());
+    let mut quotes = BTreeMap::<u32, BTreeMap<Address, TokenCfg>>::new();
+    for token in tokens.iter().filter(|token| {
+        extension_bool(token, MARKET_QUOTE_EXTENSION)
+            && selected_chains
+                .as_ref()
+                .is_none_or(|chains| chains.contains(&token.network.chain_id))
+    }) {
+        quotes
+            .entry(token.network.chain_id)
+            .or_default()
+            .entry(token.address)
+            .or_insert_with(|| token.clone());
+    }
+
+    quotes
+        .into_iter()
+        .map(|(chain_id, tokens)| match tokens.into_values().collect::<Vec<_>>().as_slice() {
+            [quote] => Ok((chain_id, quote.clone())),
+            _ => Err(RaindexError::PreflightError(format!(
+                "multiple tokens with {MARKET_QUOTE_EXTENSION}=true are configured for chain {chain_id}"
+            ))),
+        })
+        .collect()
+}
+
+fn markets_from_indexed_orders(
+    orders: impl IntoIterator<Item = IndexedOrder>,
+    registry_tokens: &[TokenCfg],
+    quotes: &BTreeMap<u32, TokenCfg>,
+    options: &MarketListOptions,
+) -> Result<Vec<RaindexMarket>, RaindexError> {
+    let registry_lookup = registry_token_lookup(registry_tokens)?;
+    let mut accumulators = BTreeMap::<(u32, Address), MarketAccumulator>::new();
+
+    for order in orders {
+        let Some(quote_cfg) = quotes.get(&order.chain_id) else {
+            continue;
+        };
+        let quote_address = quote_cfg.address;
+        let quote = market_token(quote_cfg)?;
+
+        if order
+            .inputs
+            .iter()
+            .any(|token| token.address == quote_address)
+        {
+            for token in order
+                .outputs
+                .iter()
+                .filter(|token| token.address != quote_address)
+            {
+                add_market(
+                    &mut accumulators,
+                    order.chain_id,
+                    order.raindex,
+                    token,
+                    quote.clone(),
+                    &registry_lookup,
+                )?;
+            }
+        }
+        if order
+            .outputs
+            .iter()
+            .any(|token| token.address == quote_address)
+        {
+            for token in order
+                .inputs
+                .iter()
+                .filter(|token| token.address != quote_address)
+            {
+                add_market(
+                    &mut accumulators,
+                    order.chain_id,
+                    order.raindex,
+                    token,
+                    quote.clone(),
+                    &registry_lookup,
+                )?;
+            }
+        }
+    }
+
+    let selected_tickers = options
+        .ticker_ids
+        .as_ref()
+        .map(|ids| ids.iter().cloned().collect::<HashSet<_>>());
+    let mut markets = accumulators
+        .into_values()
+        .map(|accumulator| {
+            let id = format!(
+                "{}:{:#x}:{:#x}",
+                accumulator.base.chain_id, accumulator.base.address, accumulator.quote.address
+            );
+            let ticker_id = format!(
+                "{:#x}_{:#x}",
+                accumulator.base.address, accumulator.quote.address
+            );
+            RaindexMarket {
+                id,
+                ticker_id,
+                chain_id: accumulator.base.chain_id,
+                base: accumulator.base,
+                quote: accumulator.quote,
+                raindex_addresses: accumulator.raindex_addresses.into_iter().collect(),
+            }
+        })
+        .filter(|market| {
+            selected_tickers
+                .as_ref()
+                .is_none_or(|tickers| tickers.contains(&market.ticker_id))
+        })
+        .collect::<Vec<_>>();
+    markets.sort_by(|a, b| {
+        (a.chain_id, a.base.symbol.as_str(), a.base.address).cmp(&(
+            b.chain_id,
+            b.base.symbol.as_str(),
+            b.base.address,
+        ))
+    });
+    Ok(markets)
+}
+
+fn add_market(
+    markets: &mut BTreeMap<(u32, Address), MarketAccumulator>,
+    chain_id: u32,
+    raindex: Address,
+    indexed_token: &IndexedToken,
+    quote: RaindexMarketToken,
+    registry_lookup: &HashMap<(u32, Address), TokenCfg>,
+) -> Result<(), RaindexError> {
+    let base = match registry_lookup.get(&(chain_id, indexed_token.address)) {
+        Some(token) => market_token(token)?,
+        None => indexed_market_token(chain_id, indexed_token),
+    };
+    if base.address == quote.address {
+        return Ok(());
+    }
+    markets
+        .entry((chain_id, base.address))
+        .and_modify(|market| {
+            market.raindex_addresses.insert(raindex);
+        })
+        .or_insert_with(|| MarketAccumulator {
+            base,
+            quote,
+            raindex_addresses: BTreeSet::from([raindex]),
+        });
+    Ok(())
+}
+
+fn registry_token_lookup(
+    tokens: &[TokenCfg],
+) -> Result<HashMap<(u32, Address), TokenCfg>, RaindexError> {
+    let mut sorted = tokens.to_vec();
+    sorted.sort_by_key(|token| (token.network.chain_id, token.address, token.key.clone()));
+    let mut lookup = sorted
+        .iter()
+        .map(|token| ((token.network.chain_id, token.address), token.clone()))
+        .collect::<HashMap<_, _>>();
+
+    // Variant declarations intentionally override an exact-address registry entry:
+    // an indexed legacy or underlying token still belongs to the declared canonical market.
+    for token in sorted {
+        for address in [
+            extension_address(&token, "unwrappedAddress")?,
+            extension_address(&token, "legacyAddress")?,
+        ]
+        .into_iter()
+        .flatten()
+        {
+            lookup.insert((token.network.chain_id, address), token.clone());
+        }
+    }
+    Ok(lookup)
+}
+
+fn indexed_market_token(chain_id: u32, token: &IndexedToken) -> RaindexMarketToken {
+    let address = token.address;
+    RaindexMarketToken {
+        chain_id,
+        address,
+        name: token
+            .name
+            .clone()
+            .or_else(|| token.symbol.clone())
+            .unwrap_or_else(|| format!("{address:#x}")),
+        symbol: token
+            .symbol
+            .clone()
+            .or_else(|| token.name.clone())
+            .unwrap_or_else(|| format!("{address:#x}")),
+        decimals: Some(token.decimals),
+        logo_uri: None,
+        extensions: None,
+        unwrapped_address: None,
+        legacy_address: None,
+        receipt_address: None,
+        variants: vec![address],
+    }
+}
+
+fn market_token(token: &TokenCfg) -> Result<RaindexMarketToken, RaindexError> {
+    let unwrapped_address = extension_address(token, "unwrappedAddress")?;
+    let legacy_address = extension_address(token, "legacyAddress")?;
+    let receipt_address = extension_address(token, "receiptAddress")?;
+    let mut variants = [Some(token.address), unwrapped_address, legacy_address]
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>();
+    variants.sort_unstable();
+    variants.dedup();
+
+    Ok(RaindexMarketToken {
+        chain_id: token.network.chain_id,
+        address: token.address,
+        name: token.label.clone().unwrap_or_else(|| token.key.clone()),
+        symbol: token.symbol.clone().unwrap_or_else(|| token.key.clone()),
+        decimals: token.decimals,
+        logo_uri: token.logo_uri.as_ref().map(ToString::to_string),
+        extensions: token.extensions.clone(),
+        unwrapped_address,
+        legacy_address,
+        receipt_address,
+        variants,
+    })
+}
+
+fn extension_string<'a>(token: &'a TokenCfg, key: &str) -> Option<&'a str> {
+    token
+        .extensions
+        .as_ref()
+        .and_then(|extensions| extensions.get(key))
+        .and_then(Value::as_str)
+}
+
+fn extension_bool(token: &TokenCfg, key: &str) -> bool {
+    token
+        .extensions
+        .as_ref()
+        .and_then(|extensions| extensions.get(key))
+        .and_then(Value::as_bool)
+        .unwrap_or(false)
+}
+
+fn extension_address(token: &TokenCfg, key: &str) -> Result<Option<Address>, RaindexError> {
+    extension_string(token, key)
+        .map(|address| {
+            address.parse::<Address>().map_err(|error| {
+                RaindexError::PreflightError(format!(
+                    "token {} has an invalid {key}: {error}",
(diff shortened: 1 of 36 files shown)


findolor:
* **#2848** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2847** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2845** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2845).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/common/src/raindex_client/markets/catalog.rs
```diff
@@ -0,0 +1,133 @@
...
+        .extensions
+        .as_ref()
+        .and_then(|extensions| extensions.get(key))
+        .and_then(Value::as_bool)
+        .unwrap_or(false)
+}
+
+fn extension_address(token: &TokenCfg, key: &str) -> Option<Address> {
+    extension_string(token, key).and_then(|address| address.parse().ok())
+}
```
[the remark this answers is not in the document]
findolor: Fixed in 219fdf8c1: malformed token extension addresses now propagate as discovery errors, with regression coverage.

crates/common/src/raindex_client/markets/orderbook.rs
```diff
@@ -0,0 +1,445 @@
...
+            })
+            .cloned()
+            .collect::<Vec<_>>();
+        bids.sort_by(|a, b| float_cmp_desc(a.price, b.price));
+        asks.sort_by(|a, b| float_cmp_asc(a.price, b.price));
+        bids.truncate(depth / 2);
+        asks.truncate(depth / 2);
+
+        let best_bid = bids.first().map(|level| level.price);
+        let best_ask = asks.first().map(|level| level.price);
```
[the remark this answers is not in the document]
findolor: Fixed in 219fdf8c1: odd depths use ceiling division, so depth 1 retains the best level on both sides; covered by a focused test.

crates/common/src/raindex_client/markets/trades.rs
```diff
@@ -0,0 +1,262 @@
...
+    }))
+}
+
+fn normalized_trade_event_kind(kind: &str) -> &'static str {
+    if kind.eq_ignore_ascii_case("clear") {
+        "clear"
+    } else {
+        "takeOrder"
+    }
+}
```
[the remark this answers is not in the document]
findolor: Fixed in 219fdf8c1: subgraph and local trade-kind mappings are normalized explicitly and unrecognized values remain unknown; covered by tests.

packages/webapp/src/lib/components/MarketStatistics.svelte
```diff
@@ -0,0 +1,325 @@
...
+			snapshot.market.base.name.toLowerCase().includes(normalizedSearch) ||
+			snapshot.market.base.address.toLowerCase().includes(normalizedSearch) ||
+			snapshot.market.quote.symbol.toLowerCase().includes(normalizedSearch) ||
+			snapshot.market.quote.address.toLowerCase().includes(normalizedSearch);
+		const hasQuote = Boolean(snapshot.orderbook.bestBid || snapshot.orderbook.bestAsk);
+		const matchesAvailability =
+			availability === 'all' ||
+			(availability === 'quoted' && hasQuote && snapshot.errors.length === 0) ||
+			(availability === 'inactive' && !hasQuote && snapshot.stats.tradeCount24h === 0) ||
+			(availability === 'partial' && snapshot.errors.length > 0);
```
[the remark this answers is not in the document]
findolor: This UI code was removed from this bottom PR and now lives in the dedicated webapp PR #2848, so no change belongs on #2845.

crates/common/src/raindex_client/markets/trades.rs
```diff
@@ -0,0 +1,358 @@
...
+            .collect::<Vec<_>>();
+        market_trades.sort_by(|a, b| {
+            (b.trade.block_number, b.trade.trade_event_id.as_str())
+                .cmp(&(a.trade.block_number, a.trade.trade_event_id.as_str()))
+        });
+        let latest_order_is_ambiguous = market_trades
+            .first()
+            .zip(market_trades.get(1))
+            .is_some_and(|(first, second)| first.trade.block_number == second.trade.block_number);
+        let stats = stats_from_trades(&market_trades).unwrap_or_default();
```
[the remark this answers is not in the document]
findolor: Fixed in 219fdf8c1: statistics failures now add a market data error instead of silently reporting zero statistics.

crates/common/src/raindex_client/markets/orderbook.rs
```diff
@@ -0,0 +1,865 @@
...
+        #[cfg(not(target_family = "wasm"))]
+        let ratio_vaults = vaults.clone();
+        #[cfg(not(target_family = "wasm"))]
+        let response = with_market_timeout(
+            async move { erc4626::batch_share_ratios(&provider, ratio_vaults, None).await },
+            2_000,
+        )
+        .await;
+        #[cfg(target_family = "wasm")]
+        let response = Some(erc4626::batch_share_ratios(&provider, vaults.clone(), None).await);
```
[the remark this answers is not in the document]
findolor: Fixed in fe6e64720: provider construction failures are aggregated and RPC cycling continues; the regression test verifies every malformed RPC is reported.

crates/quote/src/rpc.rs
```diff
@@ -271,11 +298,64 @@ pub async fn batch_quote(
...
+            match with_rpc_timeout(
+                batch_quote_with_provider(
+                    quote_targets,
+                    &provider,
+                    block_number,
+                    counterparty,
+                    registry,
+                    chunk_size,
+                ),
+                RPC_ATTEMPT_TIMEOUT_MS,
```
ueco-jb: `with_rpc_timeout` wraps the entire `batch_quote_with_provider` run for one RPC in `RPC_ATTEMPT_TIMEOUT_MS` (3000ms), but that run quotes every chunk of 16 targets sequentially per raindex group plus any bisection roundtrips, so a book of a few hundred order pairs needs many sequential multicalls inside one 3s budget. On realistic public RPC latency the budget expires, every configured RPC is then tried and expires the same way, `batch_quote` returns `Error::TransportError`, and `fetch_book_levels` turns that into an orderbook error for every market on the chain; each RPC also repeats the chunks that already succeeded. The existing native callers (`do_quote`, `get_order_quotes`) previously had no deadline at all, so possibly worth timing one RPC request rather than the whole batch, so a large but healthy book can still complete.

crates/common/src/raindex_client/markets/orderbook.rs
```diff
@@ -0,0 +1,948 @@
...
+        match response {
+            Some(Ok(response)) => {
+                return Ok({
+                    let values = response
+                        .items
+                        .into_iter()
+                        .filter_map(|item| {
```
ueco-jb: `read_ratios` returns the first RPC whose `batch_share_ratios` call is `Ok`, and `filter_map` then drops every item whose `data` is `None` or whose `assets_display` fails to parse. So an RPC that answers the multicall but returns per item failures ends the RPC cycling with a partial `values` map, `build_variant_map` omits the affected unwrapped or legacy variant with only a warning, `normalize_trade` drops every execution in those tokens, and the snapshot reports understated `base_volume_24h` and book depth that exclude real liquidity. Possibly worth treating a response that yields no usable ratios (or fewer than requested) as a reason to try the next RPC, so a degraded endpoint does not silently shrink the stats.

crates/quote/src/rpc.rs
```diff
@@ -177,12 +189,25 @@ async fn quote_chunk_with_probe_and_split(
         Err(err) => err,
     };
 
+    // Retrying a transport failure with progressively smaller chunks
+    // multiplies the same unhealthy RPC request and can turn a single rate
+    // limit into a request storm. Return it so the caller can cycle RPCs.
+    if matches!(initial_err, Error::TransportError(_)) {
```
agryaznov, line 209: [high] Behavior change for all existing `batch_quote` / `get_order_quotes` consumers, not only the new markets code: a transport failure used to surface as per-target `CorruptReturnData` results (callers still got a result vector), now it aborts the whole batch with `Error::TransportError` after RPC cycling.
The stopping-the-bisection-storm rationale makes sense, but the PR description does not mention this semantics change. Can you confirm the downstream callers (webapp quotes, bots) are fine losing partial results here?

crates/common/src/raindex_client/markets/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+#[cfg(not(target_family = "wasm"))]
+const DEFAULT_ORDERBOOK_DEPTH: u16 = 100;
+#[cfg(not(target_family = "wasm"))]
+const DEFAULT_RECENT_TRADES_LIMIT: u16 = 20;
+#[cfg(not(target_family = "wasm"))]
+const RATIO_READ_TIMEOUT_MS: u64 = 8_000;
+#[cfg(not(target_family = "wasm"))]
+const TRADES_READ_TIMEOUT_MS: u64 = 8_000;
+#[cfg(not(target_family = "wasm"))]
+const ORDERBOOK_READ_TIMEOUT_MS: u64 = 12_000;
```
agryaznov: [medium]  These fixed outer budgets look too tight for what runs inside them:

- `ORDERBOOK_READ_TIMEOUT_MS` (12s) wraps `fetch_book_levels`, which does two  paginated order fetches (up to `MAX_QUERY_PAGES`) plus `batch_quote`, which itself cycles RPCs at up to 3s each. Two or three slow RPCs already eat the whole budget.
- `RATIO_READ_TIMEOUT_MS` (8s) wraps `read_ratios`, which tries each RPC with  a 2s inner timeout — with more than 4 configured RPCs the later ones can never be reached.

Consider deriving the outer budget from the RPC count (or dropping the outer
timeout where the inner per-attempt timeouts already bound the work).

crates/common/src/raindex_client/markets/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+                    },
+                )
+            })
+            .collect::<BTreeMap<_, _>>();
+
+        let chain_ids = markets
+            .iter()
+            .map(|market| market.chain_id)
+            .collect::<HashSet<_>>();
+        for chain_id in chain_ids {
```
agryaznov: [medium] Chains are processed sequentially, and each chain can spend up to ~28s in
timeouts alone (ratios 8s + trades 8s + orderbook 12s). A snapshot across,
say, 7 networks can take minutes in the worst case. The per-chain work is
independent — `join_all` over the chains would cap the wall time at one
chain's worst case.

Fix: have each chain task build and return its own snapshot updates, run them
with `join_all`, and merge the results into `snapshots` afterwards (the shared
`&mut BTreeMap` can't be borrowed by concurrent futures as-is).

crates/quote/src/rpc.rs
findolor: Fixed in 6b5e3a94d: the 3-second timeout now applies to each RPC request rather than the entire multi-chunk batch. A regression test covers a healthy two-chunk batch whose aggregate duration exceeds three seconds.

crates/common/src/raindex_client/markets/orderbook.rs
findolor: Fixed in 6b5e3a94d: incomplete per-item ratio responses now continue to later RPCs, while the most complete partial response is retained only as a nonfatal fallback when no RPC can supply every ratio.

crates/quote/src/rpc.rs
findolor: Fixed in 6b5e3a94d: after exhausting RPC transport failover, batch_quote again returns an aligned per-target result vector with CorruptReturnData entries. This preserves the established caller contract without reintroducing recursive bisection on transport failures.

crates/common/src/raindex_client/markets/mod.rs
findolor: Fixed in 6b5e3a94d: the conflicting ratio and orderbook outer budgets were removed. Their individual RPC requests remain bounded by per-attempt timeouts, so every configured endpoint remains reachable.

crates/common/src/raindex_client/markets/mod.rs
findolor: Fixed in 6b5e3a94d: independent chains now build their own snapshot maps concurrently with join_all and merge deterministically afterward.

crates/common/src/raindex_client/markets/trades.rs
```diff
@@ -0,0 +1,518 @@
...
+                Err(error) => {
+                    tracing::warn!(error = %error, "excluding invalid trade from market snapshot");
+                    None
+                }
+            })
+            .collect(),
+    )
+}
+
+fn market_trade_filter(
```
agryaznov: [medium] `inputs` and `outputs` are both the full candidate set (all bases + quote), so this filter also matches base<->base trades that `normalize_trade` then throws away. On a chain with many markets this inflates the one 24h window read, which still has to fit `TRADES_READ_TIMEOUT_MS`. `latest_market_trade_filter` in get_all.rs already has the tighter shape (or-of-two-directions with the quote pinned) -- the window query can use the same.

crates/common/src/raindex_client/markets/orderbook.rs
```diff
@@ -0,0 +1,1013 @@
...
+        };
+        #[cfg(not(target_family = "wasm"))]
+        let ratio_vaults = vaults.clone();
+        #[cfg(not(target_family = "wasm"))]
+        let response = with_market_timeout(
+            async move { erc4626::batch_share_ratios(&provider, ratio_vaults, None).await },
+            2_000,
+        )
+        .await;
+        #[cfg(target_family = "wasm")]
```
agryaznov: [low] This module is only compiled for non-wasm (`#[cfg(not(target_family = "wasm"))] mod orderbook;` in mod.rs), so this `#[cfg(target_family = "wasm")]` branch and the two cfgs above it are dead code -- the body can be written straight-line. Also the `2_000` per-attempt timeout is a magic number; worth a named constant next to `TRADES_READ_TIMEOUT_MS`.

crates/common/src/raindex_client/markets/orderbook.rs
```diff
@@ -0,0 +1,1013 @@
...
+            midpoint: midpoint.and_then(|value| value.format().ok()),
+            bids: bids.into_iter().filter_map(format_level).collect(),
+            asks: asks.into_iter().filter_map(format_level).collect(),
+        };
+        if let Some(snapshot) = snapshots.get_mut(&market.id) {
+            snapshot.orderbook = book;
+            snapshot.block_number = block_number;
+            if best_bid
+                .zip(best_ask)
+                .is_some_and(|(bid, ask)| bid.gt(ask).unwrap_or(true))
```
agryaznov: [low] `bid.gt(ask).unwrap_or(true)` -- when the Float comparison itself errors, this reports "best bid is greater than best ask" although nothing is known about the book. Same `unwrap_or(true)` in the midpoint guard at L556. Keeping the conservative no-midpoint behavior is fine, but the error message should not claim a crossed book on a failed comparison.

crates/common/src/raindex_client/order_quotes.rs
```diff
@@ -456,20 +502,42 @@ pub async fn get_order_quotes_batch_with_injector(
...
+            )
+            .await?
+        }
+    };
 
     let flat_raindex: Vec<RaindexOrderQuote> = flat_results
         .into_iter()
         .map(RaindexOrderQuote::try_from_batch_order_quotes_response)
         .collect::<Result<Vec<_>, _>>()?;
+    if flat_raindex.len() != total_pair_count {
```
agryaznov: [low] Sharp edge: `get_order_quotes_inner` in the quote crate skips self-trade pairs (`input.token == output.token`) before checking `pair_is_selected`, so a caller that selects a self-trade pair gets a short response and this check turns it into a hard `PreflightError` for the whole batch. The market caller can't produce such a pair today, but the invariant is implicit -- filter self-trade pairs out of `selected_pairs` up front, or document that selected pairs must not be self-trades.

crates/common/src/raindex_client/markets/mod.rs
```diff
@@ -0,0 +1,559 @@
...
+    include_orderbook: bool,
+    include_ratios: bool,
+    include_trades: bool,
+}
+
+#[wasm_export]
+impl RaindexClient {
+    #[wasm_export(
+        js_name = "getMarkets",
+        return_description = "USDC markets discovered from active Raindex orders",
```
agryaznov: [nit] "USDC markets discovered..." -- the PR's key decision is that nothing USDC-specific is hardcoded and the quote token comes from `marketQuote`. Suggest "quote-token markets discovered from active Raindex orders" so the generated TS docs do not contradict that.

crates/quote/src/rpc.rs
```diff
@@ -177,12 +203,25 @@ async fn quote_chunk_with_probe_and_split(
...
```
agryaznov, line 209: [medium] Returning the transport error here discards the chunks that already succeeded in this provider attempt: `batch_quote_with_provider` propagates it with `?`, so with one configured RPC and, say, 100 targets (7 chunks), a 429 on chunk 7 makes all 100 targets come back `Err(CorruptReturnData("all quote RPCs failed..."))` even though chunks 1-6 held valid quotes. Before this change the failed chunk degraded per-target while the other chunks kept their data; downstream, `require_at_least_one_successful_quote` then turns the all-failed vector into a fatal error for the whole chain's orderbook. The anti-storm goal only needs the failed chunk not to be re-split -- keep the succeeded chunks' results and mark only the failed chunk's targets as failed (retrying just that chunk on the next RPC).

Related nit: if `rpcs` is empty the failover loop never runs and the message degenerates to `"all quote RPCs failed: "` with no causes.

crates/common/src/raindex_client/markets/catalog.rs
```diff
@@ -0,0 +1,707 @@
...
+                break;
+            }
+            fetched = fetched.saturating_add(page_orders.len());
+            total_count = result.total_count as usize;
+            orders.extend(page_orders);
+            if fetched >= total_count {
+                break;
+            }
+        }
+        if fetched < total_count {
```
agryaznov: [medium] This hard-fails the whole catalog in a reachable non-cap case: `total_count` is re-read each non-empty page, so the trigger is a later page coming back *empty* after the filtered set shrank mid-pagination (order deactivated/removed), or `orders_list` tolerating a partial per-subgraph failure while `orders_count` counted the full set -- the loop breaks with the stale larger `total_count` and `getMarkets` errors although a usable (slightly stale) set is in hand. The error text is also misattributed then: it says "stopped after 1000 pages" for a loop that broke on page 2. Treat an empty page as end-of-data and only fail when the page cap was actually exhausted.

Same loop shape in `fetch_orders_for_filter` in orderbook.rs L321-359 -- worth one shared paginated-fetch helper.

crates/common/src/raindex_client/markets/trades.rs
agryaznov: [medium] Setting `inputs` and `outputs` to the same candidate list does not mean input-AND-output here: both backends turn equal lists into an OR -- the local-DB builder emits `AND (input_token IN (...) OR output_token IN (...))`, and the subgraph path's in-memory filter is `inputs.contains(input) || outputs.contains(output)`. So this fetches every trade with *any* candidate token on *either* side (base<->base and base<->unlisted included). The subgraph path is worse: the `SgTradesListQueryFilters` conversion never sets the vault-token filters, so `trades_list_all` downloads the entire 24h trade set for those raindexes and filters in memory -- all inside `TRADES_READ_TIMEOUT_MS`. Correctness survives (`normalize_trade` drops non-quote-paired rows), but the window read should use the two-direction quote-pinned shape that `latest_market_trade_filter` in get_all.rs already has.

crates/common/src/raindex_client/markets/orderbook.rs
agryaznov: [low] This module is only compiled for non-wasm (`#[cfg(not(target_family = "wasm"))] mod orderbook;` in mod.rs), so this `#[cfg(target_family = "wasm")]` arm can never compile and the two `#[cfg(not(target_family = "wasm"))]` attrs above are redundant -- the body can be straight-line. Also the inline `2_000` is the only unnamed timeout in the subtree; worth a named constant next to `TRADES_READ_TIMEOUT_MS`.

crates/common/src/raindex_client/markets/orderbook.rs
agryaznov: [low] When the Float comparison itself errors, `unwrap_or(true)` here pushes "best bid is greater than best ask" -- an error-severity claim about data that was never compared. The midpoint guard at L556 fails safe (comparison error just suppresses the midpoint), so it's only this message that's wrong; distinguish the comparison-failure case or word the message so it doesn't assert a crossed book.

crates/common/src/raindex_client/order_quotes.rs
agryaznov, line 538: [low] The count check itself is load-bearing (without it the `flat_raindex[offset..offset+count]` regrouping below can panic), but the selected-path expectation it checks against is computed from `pairs[order_index].len()` without excluding self-trade pairs, while the quote crate skips self-trades *before* `pair_is_selected`. A selected self-trade pair -- producible only via registry misconfiguration today (`unwrappedAddress`/`legacyAddress` equal to the `marketQuote` token) -- would short the response and fail the whole batch here. Make the selected-path count skip self-trade pairs the same way the unselected path does.

crates/common/src/raindex_client/markets/mod.rs
agryaznov: [nit] "USDC markets discovered..." -- the PR's key decision is that nothing USDC-specific is hardcoded and the quote token comes from `marketQuote` (the only other USDC mentions in the diff are test fixtures). Suggest "quote-token markets discovered from active Raindex orders" so the generated TS docs do not contradict that.

crates/quote/src/rpc.rs
findolor: Fixed. Native quote failover now retries only the failed chunk on the next RPC, preserves completed chunks, and returns aligned failures only for a chunk that exhausts every RPC. Added regression coverage for chunk-local failover and empty RPC configuration.

crates/common/src/raindex_client/markets/catalog.rs
findolor: Fixed. Catalog and orderbook reads now share one pagination helper. An empty page is treated as end-of-data, while the error is reserved for actually exhausting the 1000-page safety limit.

crates/common/src/raindex_client/markets/trades.rs
findolor: Fixed. The market window now issues two exact quote-pinned directional reads, and the subgraph conversion pushes input/output token constraints into its query instead of downloading the full raindex trade window. Added mapping tests for union and directional intersection semantics.

crates/common/src/raindex_client/markets/orderbook.rs
findolor: Fixed. Removed the unreachable wasm branch and redundant cfg attributes, and named the timeout RATIO_RPC_TIMEOUT_MS beside the other market timeouts.

crates/common/src/raindex_client/markets/orderbook.rs
findolor: Fixed. Crossed books and Float comparison failures now produce distinct error messages; a comparison failure no longer asserts that the book is crossed.

crates/common/src/raindex_client/order_quotes.rs
findolor: Fixed. Selected-pair counts are now derived from decoded IO pairs and apply the same self-trade exclusion and selection matching as the quote crate.

crates/common/src/raindex_client/markets/mod.rs
findolor: Fixed. The generated SDK description now says quote-token markets and contains no USDC-specific claim.

crates/quote/src/rpc.rs
```diff
@@ -271,11 +312,148 @@ pub async fn batch_quote(
...
+                    .await
+                    {
+                        Ok(results) => {
+                            chunk_results = Some(results);
+                            break;
+                        }
+                        Err(Error::TransportError(error)) => {
+                            failures.push(format!("{rpc}: {error}"));
+                        }
+                        Err(error) => return Err(error),
```
ueco-jb: The loop over each chunk advances to the next entry in `providers` only on `Error::TransportError`; any other error (a provider returning `0x` maps to `Error::AlloySolTypesError`, or a multicall length mismatch) hits `Err(error) => return Err(error)`, which `group_result?` propagates and ends the whole batch before the remaining healthy RPCs are tried. `fetch_book_levels` then propagates it with `?` and `populate_chain_snapshots` publishes an empty book for every market on the chain. Possibly worth noting that here, so that a decode error from one provider cycles like a transport failure instead of aborting.

agryaznov (approved):
LGTM

PR #2847, opened 2026-08-25
feat: replace REST API with Raindex market service

findolor:
## Chained PRs

- Depends on #2845, which adds active-order market discovery and the shared SDK read model.
- Followed by #2848, which makes the webapp consume this service for the Raindex market view.

## Summary

- Replace the authenticated take-orders REST API with a public, unauthenticated Raindex market-data service.
- Serve public ticker and executable orderbook endpoints plus the SDK snapshot shape used by the webapp.
- Back reads with the SDK persistent local indexer, a minutely last-good overview cache, and per-market executable-book caches.
- Add the production deployment architecture: OpenTofu-managed DigitalOcean resources, encrypted state, declarative NixOS, deploy-rs profiles, persistent storage, nginx/TLS, bounded logs, and verified GitHub deployments.

## Key decisions

- Wait for every configured local index to become ready before warming the first snapshot. RAINDEX_LOCAL_DB_READY_TIMEOUT_SECONDS controls the startup deadline and defaults to 600 seconds.
- GET /tickers publishes required 24-hour fields for every active direct quote market without quoting every order on each minute refresh.
- GET /orderbook?ticker_id=...&depth=... quotes only the requested market and exposes up to a fixed 1,000-level SDK snapshot. Failed RPC quotes or ratio normalization return retryable 503 responses and are not cached.
- GET /v1/markets serves lightweight cached SDK snapshots for the webapp. Filtering by ticker_id adds executable depth and bounded recent trades.
- Keep the optional historical-trades endpoint deferred until the SDK supports arbitrary indexed ranges without silently truncating its 24-hour snapshot.
- Enforce global and per-socket-IP rate limits, request IDs, structured errors, CORS, tracing, and readiness/freshness health checks.
- Emit JSON logs to stdout and daily rotated files with a 14-file retention limit; production also bounds journald to 14 days and 512 MB.
- Treat no discovered markets as a valid, cacheable snapshot: /tickers and /v1/markets return 200 []. Duplicate cross-chain ticker IDs, fatal registry/trade reads, and incomplete executable books remain errors.
- Use reproducible Crane builds, nixos-anywhere bootstrap, a pinned SSH host key, encrypted Terraform variables/state, a persistent DigitalOcean volume, declarative firewall/nginx/systemd configuration, and atomic deploy-rs rollbacks.

## Validation

- cargo test -p raindex_rest_api --locked: 26 passed
- cargo clippy -p raindex_rest_api --all-targets --locked -- -D warnings
- nix build .#raindex-api --no-link
- nix flake check --no-build --impure
- nix shell .#opentofu -c tofu -chdir=infra validate
- nix shell .#actionlint -c actionlint .github/workflows/raindex-api.yaml .github/workflows/deploy-raindex-api.yaml
- Local smoke test waited for the Base index, discovered 48 active direct quote markets from local_db, and served healthy overview, ticker, detail, and orderbook responses.

## Production prerequisites

- Provision once with the documented tfEditVars, tfPlan, tfApply, and bootstrap commands, then commit the encrypted variables/state and verified host key.
- Point the production hostname at the reserved IP. The default is api.raindex.finance; RAINDEX_API_HOSTNAME overrides it.
- Configure the Production environment with RAINDEX_API_SSH_KEY and, optionally, CACHIX_AUTH_TOKEN and RAINDEX_API_HOSTNAME.


findolor:
* **#2848** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2848?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2847** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2845** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2845?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2847).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor:
@coderabbitai full review

crates/rest_api/src/cache.rs
```diff
@@ -0,0 +1,220 @@
...
+        .any(|ticker_id| !ticker_ids.insert(ticker_id))
+    {
+        return Err(MarketCacheError::InvalidSnapshot(
+            "ticker IDs must be unique across configured chains".into(),
+        ));
+    }
+    Ok(snapshots)
+}
+
+pub(crate) fn validate_detail(
```
ueco-jb, line 201: `validate_overview` builds a `HashSet` of `ticker_id` and rejects duplicates, but `validate_detail` has no equivalent check, and `ticker_id` is built with no chain component in the SDK (`markets/catalog.rs` formats it as `{base}_{quote}` while `id` carries the chain). So if the same base and quote token addresses are active on two chains, `get_market` returns both snapshots and `CachedMarketData::market` takes `details.first()`, so `/v1/markets?ticker_id=` and `/orderbook?ticker_id=` serve whichever chain the SDK ordered first while `/tickers` refuses the pair as a duplicate. Possibly worth either qualifying `ticker_id` by chain or applying the same duplicate check on the detail path, so the two endpoints cannot disagree about which chain a ticker means.

findolor: Fixed in f6dd9f737: detail snapshots now enforce the same case-insensitive ticker uniqueness check as overview snapshots. Ambiguous cross-chain pairs are rejected explicitly instead of silently selecting the first chain; regression coverage was added.

crates/rest_api/src/cache.rs
```diff
@@ -0,0 +1,226 @@
...
+            last_success_at: state.last_success_at,
+            refresh_healthy: state.refresh_healthy,
+        }
+    }
+
+    fn overview_options(&self) -> MarketSnapshotOptions {
+        MarketSnapshotOptions {
+            orderbook_depth: Some(self.orderbook_depth),
+            recent_trades_limit: Some(self.recent_trades_limit),
+            include_orderbook: Some(false),
```
agryaznov: [medium] `include_orderbook: Some(false)` means every overview snapshot keeps `RaindexMarketOrderbook::default()` -- so in production `/tickers` can never emit `bid`/`ask` (both are `skip_serializing_if` in `TickerResponse`, sourced from `snapshot.orderbook.best_bid/best_ask`), and `/v1/markets` without a ticker filter always returns empty `bids`/`asks` despite the README calling it the complete overview. The route test that asserts `bid == Some("1.9")` passes only because `MockMarketData` injects a book the real overview path cannot produce. Either compute best bid/ask in the overview refresh, or drop the fields from the ticker payload and document it.

crates/rest_api/src/service.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    async fn market(
+        &self,
+        ticker_id: &str,
+    ) -> Result<Option<Arc<RaindexMarketSnapshot>>, MarketDataError> {
+        let overview = self.snapshots().await?;
+        let details = self
+            .cache
+            .get_market(&self.provider, ticker_id)
```
agryaznov, line 162: [medium] Any request with an unknown `ticker_id` reaches the SDK, and `get_market_snapshots` runs full market discovery (two filtered paged active-order reads per configured chain) before the ticker filter is applied -- the empty result is then cached per unique ticker string, so varying the string gets one full discovery read per request within the rate limits (local-DB/subgraph load; RPC work is skipped when zero markets match). `market()` already holds `overview` -- return `Ok(None)` early when the ticker is not in it, so garbage 404s without touching the SDK. That early check also fixes the zero-stats fallback below.

crates/rest_api/src/cache.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    pub async fn get_market(
+        &self,
+        provider: &RaindexProvider,
+        ticker_id: &str,
+    ) -> Result<Arc<Vec<RaindexMarketSnapshot>>, Arc<MarketCacheError>> {
+        let key = ticker_id.to_ascii_lowercase();
+        let options = self.market_options(&key);
+        self.details
+            .try_get_with(key, async move {
```
agryaznov: [medium] `try_get_with` caches only `Ok` results, and one error-severity `ratios`/`orderbook` entry makes `validate_detail` fail the whole result -- so while an RPC fault lasts, every `/orderbook?ticker_id=X` and `/v1/markets?ticker_id=X` redoes full discovery plus the ratio read and book quotes against the same failing RPC (up to 120 rpm/IP, 6000 rpm global), returning 503s the entire time. "Failed quotes are not cached" is the right contract for freshness, but it needs a short negative-cache TTL (a few seconds) or serve-last-good-with-staleness-marker so a transient fault does not become a request-driven RPC stampede.

crates/rest_api/src/routes/markets.rs
```diff
@@ -0,0 +1,560 @@
...
+pub async fn orderbook(
+    ticker_id: &str,
+    depth: Option<&str>,
+    _limit: PublicRateLimit,
+    state: &State<AppState>,
+) -> Result<Json<OrderbookResponse>, ApiError> {
+    let depth = parse_depth(depth)?;
+    let ticker_id = canonical_ticker_id(ticker_id);
+    let snapshot = market_snapshot(state, &ticker_id).await?;
+    let per_side = if depth == 0 { usize::MAX } else { depth / 2 };
```
agryaznov: [low] `depth / 2` truncates odd depths down: `depth=1` (accepted by `validate_depth`) returns a 200 with empty `bids` and `asks`, and `depth=3` drops a level. The SDK's own truncation uses `depth.div_ceil(2)` for exactly this reason, and since the cache always requests the full 1000-level snapshot, this division is the only truncation the caller's `depth` reaches. Use `div_ceil(2)` here too.

crates/rest_api/src/service.rs
```diff
@@ -0,0 +1,440 @@
...
+            .await
+            .map_err(|error| MarketDataError::Unavailable(error.to_string()))?;
+        Ok(details.first().map(|detail| {
+            Arc::new(
+                overview
+                    .iter()
+                    .find(|snapshot| snapshot.market.id == detail.market.id)
+                    .map(|snapshot| merge_detail(snapshot, detail))
+                    .unwrap_or_else(|| detail.clone()),
```
agryaznov: [low] When the market is missing from the last overview (appeared after the refresh, or refresh is stale-on-error), this fallback serves the detail snapshot fetched with `include_trades: false` -- i.e. `RaindexMarketStats::default()` zeros and an empty trade list presented as real data on `/v1/markets?ticker_id=...` until the next successful refresh. Related: on the successful merge path, `merged.observed_at = detail.observed_at` stamps up-to-TTL-old overview stats and trades with the fresh detail timestamp. The early overview-membership 404 suggested above removes the fallback case entirely.

.github/workflows/deploy-raindex-api.yaml
```diff
@@ -0,0 +1,88 @@
...
+            system) nix run .#deployNixos -- -i ~/.ssh/id_ed25519 ;;
+            all) nix run .#deployAll -- -i ~/.ssh/id_ed25519 ;;
+            *) echo "Unsupported deployment scope: $SCOPE" >&2; exit 1 ;;
+          esac
+      - name: Verify deployment
+        run: |
+          api_hostname="${RAINDEX_API_HOSTNAME:-api.raindex.finance}"
+          health_url="https://$api_hostname/health/detailed"
+          for _ in $(seq 1 60); do
+            if curl --fail --silent --show-error --max-time 5 "$health_url"; then
```
agryaznov: [low] `curl --fail` on `/health/detailed` exits 0 on any 200, and the route returns 200 with `status: "degraded"` when sync or the snapshot cache is unhealthy -- so the verify step proves reachability, not health. The window is narrow (startup aborts unless warm-up succeeded, so degraded here means a post-start regression), but the check is one pipe away from meaning what it says: `curl ... | jq -e '.status == "ok"'`.

crates/rest_api/src/cache.rs
findolor: Fixed. Overview snapshots now request depth 2 with orderbook reads enabled, which supplies one executable best level per side for tickers and overview responses without loading full depth. Added an options regression test.

crates/rest_api/src/service.rs
findolor: Fixed. market() now checks the cached overview first and returns None for an unknown ticker before invoking the SDK detail read.

crates/rest_api/src/cache.rs
findolor: Fixed. Detail failures now use a separate five-second negative cache, preventing transient orderbook or ratio faults from causing request-driven RPC stampedes while keeping the failure cache short-lived.

crates/rest_api/src/routes/markets.rs
findolor: Fixed. Odd depth values now use div_ceil(2), with route coverage confirming depth 3 returns two levels per side.

crates/rest_api/src/service.rs
findolor: Fixed. Overview membership is now required before detail reads, the zero-stat detail fallback is gone, and merge_detail preserves the overview observation timestamp for its cached statistics.

.github/workflows/deploy-raindex-api.yaml
findolor: Fixed. Deployment verification now pipes detailed health through jq and succeeds only when status equals ok.

infra/default.nix
```diff
@@ -0,0 +1,168 @@
...
+    '';
+  };
+
+  tfPlan = rainix.mkTask.${system} {
+    name = "tf-plan";
+    additionalBuildInputs = buildInputs;
+    body = ''
+      ${preamble}
+      ${decryptState}
+      tofu -chdir=infra plan -out=tfplan "$@"
```
ueco-jb: `tfPlan` writes `infra/tfplan` with `plan -out=tfplan`, and an OpenTofu plan file stores the input variable values (including `do_token`) in cleartext, but `tfPlan` uses `preamble`, whose `on_exit` runs `cleanup` (removing `tfState`/`tfVars`/`tfSecretVars`, not `tfPlanFile`; only `cleanupWithPlan`, used by `tfApply`, removes it). `infra/.gitignore` ignores `*.tfplan`, which does not match the extensionless `tfplan`, so a `git add -A` after a run that plans without applying stages a plaintext file holding the DigitalOcean token. Possibly worth noting that here, so that a run that only plans removes `tfPlanFile`, or `.gitignore` also covers `tfplan` (following infra/.gitignore:5).

crates/rest_api/src/cache.rs
```diff
@@ -0,0 +1,255 @@
...
+
+pub(crate) fn validate_overview(
+    snapshots: Vec<RaindexMarketSnapshot>,
+) -> Result<Vec<RaindexMarketSnapshot>, MarketCacheError> {
+    if snapshots
+        .iter()
+        .flat_map(|snapshot| &snapshot.errors)
+        .any(|error| {
+            error.severity == RaindexMarketDataErrorSeverity::Error
+                && (error.source == "trades" || error.source == "registry")
```
ueco-jb: `validate_overview` only rejects `source` values `"trades"` and `"registry"`, so when `populate_chain_snapshots` pushes an `Error` severity entry with `source` `"ratios"` or `"orderbook"` (a ratio read that failed across the chain, or a crossed book) the snapshot still passes and `refresh_overview` overwrites `state.snapshots` with it and sets `refresh_healthy = true`. `/tickers` then serves understated volumes and a crossed `best_bid`/`best_ask` that `validate_detail` refuses with 503, while `/health/detailed` reports healthy and the `retaining the previous snapshot` path never runs. Possibly worth noting that here, so that `validate_overview` and `validate_detail` treat the same `orderbook` and `ratios` errors alike.

agryaznov (approved):
LGTM

PR #2848, opened 2026-08-25
feat: show Raindex market data

findolor:
## Chained PRs

- Depends on #2847, which provides the public cached market API.
- #2847 depends on #2845, which provides active-order discovery and the shared SDK market read model.

## Summary

- Add a dedicated /markets Raindex market page that consumes only the public REST market service.
- Show every active direct quote-token pair discovered from Raindex orders, with API-backed 24-hour statistics, token artwork when available, and on-demand executable depth.
- Source market content from the public REST service while retaining the application shared layout and initialization behavior.

## Key decisions

- Present the route as a Raindex-level market view and link it immediately after Vaults in permanent navigation.
- Refresh the cached overview each minute and quote only the selected market. TanStack Query cancellation prevents superseded requests from accumulating.
- Render registry-provided token images when available and retain token initials as the missing-image fallback.
- Treat 200 [] as healthy: counters render as zero and the page explains that no active Raindex orders currently provide a direct pair with the configured quote token.
- Preserve the last successful overview or detail when a background refresh fails and surface a compact freshness status.
- Keep scanning values concise while exposing unrounded API decimal strings in selected-market verification details.
- Support search by symbol, token name/address, ticker ID, quote symbol, and chain ID without product category, ISIN, or market-enabled hardcoding.

## Validation

- npm run check: 0 errors and 0 warnings
- npm run lint
- Focused market/navigation/component tests: 8 passed
- Full webapp suite: 32 files and 224 runnable tests passed; the existing network-dependent fullDeployment.test.ts could not fetch its remote registry and skipped 9 tests
- npm run build
- git diff --check
- Manual local integration against the local API and index: 48 active markets rendered, including non-stock tokens; search, refresh, token artwork, empty executable depth, populated depth, exact values, and recent trades were verified.

Build output retains the repository existing sourcemap, large-chunk, and optional dependency warnings.


findolor:
* **#2848** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2847** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2845** <a href="https://app.graphite.com/github/pr/rainlanguage/raindex/2845?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/rainlanguage?ref=merge-queue-instructions-in-stack-comment&prId=2848).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor:
## Updated Raindex market UI

Captured from the current Vercel preview: [open the live markets page](https://rain-orderbook-v6-8i4wo8oah-rain-x-h20.vercel.app/markets).

### Desktop overview

Raindex market overview

### Executable market detail

Raindex executable market orderbook and recent trades

### Exact API values

Raindex exact market API values

### Search across active-order tokens

The example below filters the Raindex-wide market list to WETH; discovery is not limited to stock tokens.

Raindex market search filtered to WETH

### Mobile overview

Raindex mobile market overview

### Mobile executable detail

Raindex mobile executable market detail

packages/webapp/src/routes/markets/+page.svelte
```diff
@@ -0,0 +1,153 @@
+<script lang="ts">
```
ueco-jb, line 1: The PR body says `/markets` returns before importing or initializing the browser SDK and SQLite WASM runtime, but this route renders under the root `+layout.ts`, which statically imports `DotrainRegistry` and `SQLiteWasmDatabase` plus `init` and whose load awaits the registry fetch, the WASM init, and `getRaindexClient` before any child page renders. So when the registry endpoint is down the shared layout shows its error page and `/markets` never calls the market API, the opposite of the claimed independence from the SQLite indexer and registry. Possibly worth a route group or layout reset that renders `/markets` without the root layout load, so the page depends only on the public market service as described.

findolor: The shared layout initialization is intentional: the product requirement is for Markets to retain the normal application layout and WASM initialization. I corrected the PR description so it no longer claims this route bypasses those dependencies.

packages/webapp/src/lib/market-api/client.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type MarketApiErrorBody = {
+  request_id?: string;
+  error?: {
+    code?: string;
+    message?: string;
+  };
+};
+
+export const DEFAULT_MARKET_API_URL = "http://127.0.0.1:8000";
```
agryaznov: [medium] `http://127.0.0.1:8000` is the fallback whenever `PUBLIC_MARKET_DATA_API_URL` is unset -- and nothing in the repo sets it (the Vercel workflows map only `WALLETCONNECT_PROJECT_ID`), so the first production deploy serves /markets fetching localhost, blocked as mixed content from an https page, and every visitor sees the service-unavailable alert until someone sets the var in Vercel. The existing `PUBLIC_WALLETCONNECT_PROJECT_ID` avoids this class of miss by going through `$env/static/public`, which fails the build when unset. Either default to the service's production URL (`https://api.raindex.finance`, the sibling PR's default hostname), or give this var the same build-time guard.

packages/webapp/src/routes/markets/+page.svelte
```diff
@@ -0,0 +1,153 @@
...
+			<div class="space-y-3">
+				{#each [0, 1, 2, 3, 4, 5, 6] as skeleton (skeleton)}
+					<div class="h-14 animate-pulse rounded-md bg-gray-100 dark:bg-gray-800"></div>
+				{/each}
+			</div>
+		</div>
+	{:else if $marketsQuery.isError && markets.length === 0}
+		<div class="rounded-lg border border-red-200 bg-red-50 p-6 text-center dark:border-red-900/60 dark:bg-red-950/20" role="alert">
+			<h2 class="font-semibold text-gray-900 dark:text-white">Market service is unavailable</h2>
+			<p class="mt-1 text-sm text-gray-600 dark:text-gray-400">The cached market overview could not be loaded. Existing order, vault, and trade pages are unaffected.</p>
```
agryaznov: [medium] This copy claims independence the route does not have in the other direction: the root layout's `load` sets `errorMessage` when the `DotrainRegistry` fetch (raw.githubusercontent.com) fails, and `+layout.svelte` renders `<ErrorPage />` for every pathname except `/` -- so a registry outage takes down /markets before it issues a single market-API request, even though this page needs neither the registry nor the WASM DB. Keeping the shared layout initialization is fine per the earlier thread; the ask is one line narrower: exempt `/markets` from the `errorMessage` gate the way `/` is exempt, so the page degrades to its own error handling instead of the app-wide one.

packages/webapp/src/routes/markets/page.test.ts
```diff
@@ -0,0 +1,53 @@
...
+	})
+}));
+
+describe('Raindex market data page', () => {
+	it('treats an empty active market set as a healthy state', () => {
+		render(Page);
+
+		expect(screen.getByRole('heading', { name: 'Raindex market data' })).toBeInTheDocument();
+		expect(
+			screen.queryByText(
```
agryaznov: [nit] This negative assertion checks for a sentence ("Direct quote-token pairs discovered...") that no component in the PR renders -- it exists only here, so the assertion can never fail. Looks like a leftover from a removed subtitle; drop it (the neighbouring `Market service is unavailable` absence check is the meaningful one).

packages/webapp/src/lib/market-api/format.ts
```diff
@@ -0,0 +1,39 @@
...
+  const number = Number(value);
+  return Number.isFinite(number)
+    ? new Intl.NumberFormat("en-US", {
+        notation: "compact",
+        maximumSignificantDigits: 6,
+      }).format(number)
+    : value;
+}
+
+export function formatSpread(
```
agryaznov, line 22: [nit] No sign handling: on a crossed book (the SDK does serve `bestBid > bestAsk`, with an error entry the panel shows in the amber block) this prints a negative spread, and `toFixed(2)` on a tiny negative renders `-0.00%`. Consider clamping to `—` or labelling "crossed" when `ask < bid` so the spread cell doesn't report a nonsensical negative percentage.

packages/webapp/src/lib/market-api/client.ts
findolor: Fixed. The client now defaults to https://api.raindex.finance when no public override is configured, with a regression test for the production-safe fallback.

packages/webapp/src/routes/markets/+page.svelte
findolor: Fixed. The markets route is exempt from the registry error-page gate while retaining the normal shared providers and layout, so its API-owned error handling remains available during a registry outage. Added layout coverage.

packages/webapp/src/routes/markets/page.test.ts
findolor: Fixed. Removed the ineffective assertion for copy that is not rendered.

packages/webapp/src/lib/market-api/format.ts
findolor: Fixed. Crossed or non-finite bid/ask inputs now render an em dash rather than a negative spread, with crossed-book test coverage.

packages/webapp/src/routes/markets/+page.svelte
```diff
@@ -0,0 +1,153 @@
...
+				market.tickerId,
+				String(market.chainId)
+			].some((value) => value.toLowerCase().includes(normalizedSearch));
+		})
+		.sort(
+			(a, b) =>
+				Number(Boolean(b.stats.lastPrice)) - Number(Boolean(a.stats.lastPrice)) ||
+				a.market.base.symbol.localeCompare(b.market.base.symbol)
+		);
+	$: activeMarketCount = markets.filter((snapshot) => snapshot.stats.lastPrice).length;
```
ueco-jb: `activeMarketCount` counts markets by `stats.lastPrice` presence, but the SDK `stats_from_trades` sets `last_price` from the newest trade even when `trade_count_24h` is `0` (test `historical_last_price_does_not_change_24_hour_aggregates`). That count renders under the `Traded 24h` tile, so a market last traded days ago is counted as traded in the last 24 hours. Possibly worth noting that here, so that the tile keys on `stats.tradeCount24h > 0` instead (following +page.svelte:60-61).

agryaznov (approved):
LGTM

PR #2877, opened 2026-09-16
Bump rain.strategies registry for Robinhood local DB sync

findolor:
## Dependent PR

- https://github.com/rainlanguage/rain.strategies/pull/132

## Motivation

The web application and REST API need to consume the requested rain.strategies registry revision, which enables Robinhood local DB sync.

## Solution

- Pin the web application registry URL to `749c0becb5ec8b1e403cebf4b3edffb9ea4908be`.
- Pin the REST API default registry URL to the same revision.

## Checks

- [x] `cargo test -p raindex_rest_api config::tests`
- [x] `npm run check -w @rainlanguage/webapp` — 0 errors and 0 warnings
- [x] Requested raw registry URL returns HTTP 200
- [x] Commit hooks: no-consumer-prettier, prettier-rainix, and rustfmt passed

## Summary by CodeRabbit

* **Updates**
  * Updated the default registry reference to a newer registry revision.
  * Updated the web application to use the same registry revision for consistency.


Siddharth2207:
## Review

**Verdict:** Approve the pin bump. The PR itself looks correct.

### Diff check
- Webapp `REGISTRY_URL` and REST API `DEFAULT_REGISTRY_URL` both point at `749c0becb5ec8b1e403cebf4b3edffb9ea4908be`.
- That registry revision returns HTTP 200 and only adds `local-db-sync.robinhood` in settings (base config unchanged).

### Follow-up (pre-existing, not introduced here)

`LocalDbSyncGate` non-blocking notices / `synccomplete` only fire while `localDbSyncGate.status` is `syncing` or `failure`. After the first successful local DB sync the gate stays `ready`, so a newly enabled network (Robinhood via this bump) can sync or fail without a banner or list refresh for users who already completed initial sync.

Worth a separate fix if we want Robinhood local DB progress to surface for returning users — shouldn't block merging this registry pin.

Siddharth2207 (approved):

